PR #127, opened 2024-09-19
Update Examples

Mattjet27:

Updates the examples to reflect latest updates.

It helps to ensure the examples are kept up to date ahead of the next release.

Tested the newly added Pattern Triggered Measurements using physical hardware (STS M2 T2 tester).


lornazh (approved):

Examples/source/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -57,6 +144,132 @@ internal static void PinSiteDataMethods()
...
         }
 
+        internal static void ConstructWithArrays()
+        {
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VCC1", "VCC2" };
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 0 };
+            // Per-pin SiteData objects.
+            var pinPinSiteData = new[]
```
lornazh, line 154: Variable name should be p**er**PinSiteData.

Examples/source/InstrumentAbstraction/ForceVoltagePPMU.cs
```diff
@@ -17,75 +23,128 @@ namespace NationalInstruments.SemiconductorTestLibrary.Examples.InstrumentAbstra
...
 
         /// <summary>
         /// This example demonstrates how to configure different voltage levels for each site.
-        /// The example assumes that there are 2 pins and 4 sites.
+        /// The example assumes there is a maximum of 4 sites.
         /// </summary>
         /// <param name="tsmContext">The <see cref="ISemiconductorModuleContext"/> object.</param>
-        internal static void DifferentValuesPerSitePpmu(ISemiconductorModuleContext tsmContext)
+        /// <param name="ppmuPinNames">The PPMU pins to force voltage on</param>
+        internal static void DifferentValuesPerSitePmmu(ISemiconductorModuleContext tsmContext, string[] ppmuPinNames)
```
lornazh, line 88: DifferentValuesPerSiteP**p**mu. And maybe also add Pins in the end, so DifferentValuesPerSitePpmuPins?

Examples/source/InstrumentAbstraction/ForceVoltageSMU.cs
```diff
@@ -19,88 +20,111 @@ namespace NationalInstruments.SemiconductorTestLibrary.Examples.InstrumentAbstra
...
 
         /// <summary>
         /// This example demonstrates how to configure different voltage levels for each site.
-        /// The example assumes there are 2 pins and 4 sites.
+        /// The example assumes there is a maximum of 4 sites.
         /// </summary>
         /// <param name="tsmContext">The <see cref="ISemiconductorModuleContext"/> object.</param>
-        internal static void DifferentValuesPerSiteSmu(ISemiconductorModuleContext tsmContext)
+        /// <param name="smuPinNames">The SMU pins to force voltage on</param>
+        internal static void DifferentValuesPerSiteSmu(ISemiconductorModuleContext tsmContext, string[] smuPinNames)
```
lornazh, line 85: DifferentValuesPerSiteSmu**Pins**?

Examples/source/InstrumentAbstraction/ForceVoltageSMU.cs
```diff
@@ -19,88 +20,111 @@ namespace NationalInstruments.SemiconductorTestLibrary.Examples.InstrumentAbstra
...
-            var perSiteVoltageLevelsArray = new double[] { 3.2, 3.3, 3.1, 3.8 };
             var sessionManager = new TSMSessionManager(tsmContext);
-            var smuPins = sessionManager.DCPower(pinNames);
-
-            var voltageLevelsSiteData = new SiteData<double>(perSiteVoltageLevelsArray);
+            // Max Site count of 4 expected
+            var perSiteVoltageLevelsArray = new double[] { 3.2, 3.3, 3.1, 3.8 };
+            var activeSites = tsmContext.SiteNumbers.ToArray();
+            // Array must be resized based on number of on active sites
+            Array.Resize(ref perSiteVoltageLevelsArray, activeSites.Length);
```
lornazh, line 92: If the activeSites is {1, 3}, would the resized array be {3.2, 3.3} or {3.3, 3.8}?

Examples/source/InstrumentAbstraction/ForceVoltagePPMU.cs
Mattjet27: @lornazh, I will go with `DifferentValuesPerSiteAcrossAllPpmuPins`.

Examples/source/InstrumentAbstraction/ForceVoltageSMU.cs
Mattjet27: @lornazh, good point, I am going to change this to the following, to be clearer:

```C#
var activeSites = tsmContext.SiteNumbers.ToArray();
// Unique per-site data, assuming a max site count of 4.
// Each index of the array represents the corresponding site number.
var perSiteVoltageLevelsArray = new double[] { 3.2, 3.3, 3.1, 3.8 };
// Note the site unique values must be filtered based on the active sites.
var activeSiteUniqueValues = new double[activeSites.Length];
for (int i = activeSites.Length - 1; i >= 0; i--)
{
    activeSiteUniqueValues[i] = perSiteVoltageLevelsArray[i];
}
```

I will also do the same for `DifferentValuesPerSiteAcrossAllPpmuPins` method in the ForceVoltagePPMU.cs file as well.

lornazh: @Mattjet27 when activeSites is {1, 3}, your new logic will reproduce {3.2, 3.3} right? Which, I think is the same behavior as Array.Resize? But my understanding is, we would like to produce {3.3, 3.8}, do you agree?

Below is a way how we produce {3.3, 3.8}.

var activeSiteUniqueValues = tsmContext.SiteNumbers.Select(siteNumber => perSiteVoltageLevelsArray[siteNumber]).ToArray();

Mattjet27: @lornazh, as discussed offline, I had meant for this to be the following:
```C#
for (int i = 0; i < activeSites.Length; i++)
{ 
   activeSiteUniqueValues[i] = perSiteVoltageLevelsArray[activeSites[i]];
}
```
I will raise a new PR to address this now.

PR #141, opened 2024-10-14
Fix Leakage Test Implementation

Mattjet27:

Modifies the Leakage Test implementation to now force 0V on all pins before forcing the specified voltage and measuring. Additionally, the output is disabled on each pin after measuring. When testing serially, the output should be disabled on the previous pin or pin group before forcing voltage on the next pin or pin group. The documentation for the methods has also been updated to reflect this change

The current implementation does not adequately perform a Leakage measurement, especially not so in the serial operation scenario, Since the pins are not in a controlled state before testing begins and are left high when existing the test, which is not ideal for testing most DUTs. 

New integration testes have been added and are passing. Manual testing done both offline and with actual hardware.


SemiconductorTestLibrary.TestStandSteps/source/LeakageTest.cs
```diff
@@ -67,64 +70,80 @@ public static void LeakageTest(
...
-                    () =>
-                    {
-                        dcPower?.MeasureAndPublishCurrent("Leakage", out _);
-                    },
-                    () =>
-                    {
-                        digital?.MeasureAndPublishCurrent("Leakage", out _);
-                    });
+                    // Measure current on all pins simultaneously.
+                    InvokeInParallel(
```
lornazh, line 136: I guess there is no point of making MeasureAndPublishCurrent and PowerDown/TurnOffOutput in their own InvokeInParallel calls, now that they are separately handled in serial case? Moving them right after the ForceVoltage calls (like how you do it for serial case) should have better performance because there are less code merging points.

SemiconductorTestLibrary.TestStandSteps/source/LeakageTest.cs
```diff
@@ -67,64 +70,80 @@ public static void LeakageTest(
...
                         var dcPowerPins = tsmContext.FilterPinsByInstrumentType(new string[] { pinOrPinGroup }, InstrumentTypeIdConstants.NIDCPower);
                         if (dcPowerPins.Any())
                         {
                             var dcPowerPerPinOrPinGroup = sessionManager.DCPower(dcPowerPins);
-                            dcPowerPerPinOrPinGroup.ConfigureSourceDelay(settlingTime);
-                            dcPowerPerPinOrPinGroup.ConfigureMeasureSettings(dcPowerMeasureSettings);
-                            dcPowerPerPinOrPinGroup.ForceVoltage(voltageLevel, currentLimit);
+                            dcPowerPerPinOrPinGroup.ForceVoltage(voltageLevel, currentLimit, waitForSourceCompletion: true);
+                            dcPowerPerPinOrPinGroup.MeasureAndPublishVoltage("Leakage", out _);
+                            dcPowerPerPinOrPinGroup.PowerDown();
```
lornazh, line 107: Now that we are comparing to the legacy implementation, should we
1. Force to 0V before calling PowerDown?
2. Do PowerDown for all pins at once in the end?

lornazh (approved):

SemiconductorTestLibrary.TestStandSteps/source/LeakageTest.cs
Mattjet27: @lornazh, the idea for the parallel case is to ensure that all pins are forced at the specified levels before performing the measurement. This is important for some DUTs, and ensures constancy for how pin groups are handled in the serial case.

SemiconductorTestLibrary.TestStandSteps/source/LeakageTest.cs
Mattjet27: @lornazh, good point, I will push a commit now that we reviewed together that implements this.

PR #148, opened 2024-11-04
Add support to securely run workflows from forked repositories 

Tanish-NI:
The current GitHub workflow implementation is not capable of supporting contributions from outside contributors (externally forked repositories). However, this is unintended and outside contributions should be fully supported, as is outlined in [CONTRIBUTING.md](https://github.com/ni/semi-test-library-dotnet/blob/main/CONTRIBUTING.md). 

The [Pre-Build](https://github.com/ni/semi-test-library-dotnet/blob/main/.github/workflows/Pre-Build.yml) and Pre-Test workflows on GitHub use Azure Token to call internal Azure Pipelines. This token is stored as a repository secret which cannot be accessed by PRs from forked repositories. When an external contributor raises a PR from their fork, these workflows will fail, blocking the necessary checks from running against their code.

Added 3 new workflows:-
1. pr-build
2. pr-test
3. pr-intake

The idea is to separate the PR Intake workflow from the PR Build workflow, as suggested in the [GitHub Security Lab Article](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) using the `workflow-run` event along with `pull_request`. 
But as in this case `workflow_run` event will be triggered out of the context of the `pull_request` workflow to access the AzDo token, to communicate between these workflows, GitHub APIs are used.

`pr-intake` workflow will be triggered when a PR is raised and `pr-build` & `pr-test` workflow will be triggered when `pr-intake` workflow is approved manually (as using `gh-action-testing` environment) and starts running.
`pr-build` and `pr-test` will call Azure pipelines. And `pr-intake` uses GitHub APIs to check status and conclusion of both runs. 

Pending, because of the Note mentioned in [workflow_run](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) GitHub documentation, the pr-build and pr-test workflows need to be in main to trigger them via workflow_run event. Hence, in this case, testing will be done after the PR went in the main (default branch).
image

### DIFF
--- a/.github/workflows/pr-intake.yml
+++ b/.github/workflows/pr-intake.yml
@@ -5,6 +5,7 @@ on:
         branches: 
         - main
         - 'releases/**'
+        - 'users/tanish/**'        
         paths:
         - 'Examples/**'
         - 'SemiconductorTestLibrary.Abstractions/**'
--- a/.github/workflows/pr-intake.yml
+++ b/.github/workflows/pr-intake.yml
@@ -12,7 +12,6 @@ on:
         - 'SemiconductorTestLibrary.TestStandSteps/**'
         - 'TestAssets/**'
         - '.github/**'
-    workflow_dispatch:
 
 env:
     GitHub_Repo_Name: "semi-test-library-dotnet"
--- a/.github/workflows/pr-intake.yml
+++ b/.github/workflows/pr-intake.yml
@@ -20,7 +20,7 @@ env:
 jobs:
   wait-for-pr-build-workflow:
     runs-on: ubuntu-latest
-    environment: gh-action-testing
+    # environment: gh-action-testing # Temporary commenting out the environment to test the workflow without it.
     steps:
       # Before fetching the run ID, we need to wait for the PR Build Workflow run to start as it uses in_progress type of workflow_run event.
       - name: Wait for PR Build Workflow to Start
--- a/.github/workflows/pr-intake.yml
+++ b/.github/workflows/pr-intake.yml
@@ -35,7 +35,7 @@ jobs:
                     -o runs.json)
           
           if [ $response -ne 200 ]; then
-            echo "Failed to fetch runs of Workflow - $WORKFLOW_NAME. Exiting."
+            echo "Failed to fetch id of Workflow - $WORKFLOW_NAME. Exiting."
             exit 1
           fi
           WORKFLOW_ID=$(jq -r ".workflows[] | select(.name == \"$WORKFLOW_NAME\") | .id" runs.json)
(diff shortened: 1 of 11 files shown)


.github/workflows/pr-intake.yml
```diff
@@ -0,0 +1,157 @@
...
+    PR_Number: "#${{ github.event.number }}"
+
+jobs:
+  wait-for-pr-build-and-test-workflows:
+    runs-on: ubuntu-latest
+    environment: gh-action-testing
+    steps:
+      # Before fetching the run ID, we need to wait for the PR Build and PR Test Workflow run to start as it uses in_progress type of workflow_run event.
+      - name: Wait for PR Build and PR Test Workflow to Start
+        run: sleep 10
```
Mattjet27, line 27: @Tanish-NI. are we sure this is enough time? Maybe 30 seconds is safer?

.github/workflows/pr-intake.yml
```diff
@@ -0,0 +1,157 @@
...
+      - name: Get PR Test Workflow Run ID
+        id: get_test_runid
+        run: |
+          WORKFLOW_NAME="pr-test"
+
+          WORKFLOW_ID=$(curl -s "https://api.github.com/repos/ni/${{ env.GitHub_Repo_Name }}/actions/workflows" \
+                              | jq -r ".workflows[] | select(.name == \"$WORKFLOW_NAME\") | .id")
+
+          # Fetch runs of Workflow
+          response=$(curl -s -w "%{http_code}" "https://api.github.com/repos/ni/${{ env.GitHub_Repo_Name }}/actions/workflows/$WORKFLOW_ID/runs" \
```
Mattjet27, line 61: @Tanish-NI, shouldn't the valid response check also be done in the other run steps as well, just to be safe: `Wait for Specific Build Workflow Run to Complete`, `Wait for Specific Test Workflow Run to Complete`, `Check Conclusion of Build Workflow`, and `Check Conclusion of Test Workflow`?

Mattjet27 (changes requested):
@Tanish-NI, overall looks good. I left a few specific comments to address before approving. Additionally, I am wondering if it does not make sense for this PR to also remove the other existing Pre-Build.yml and Pre-Test.yml workflow files. Since we are using lowercase naming here, I am pretty sure we will have both sets of files in the repo if this is commited as-is. Is that intentional? Note I like the idea of using lower-case and cleaning up the old files.

.github/workflows/pr-intake.yml
Tanish-NI: Sure @Mattjet27 !
Added the HTTP status check after each API call.

.github/workflows/pr-intake.yml
Tanish-NI: Yes, 30 seconds sounds safer to assume that pr-build and pr-test runs should be triggered.
Updated!

Tanish-NI:
@Mattjet27 I think we should not remove the existing Pre Build and Pre Test workflows right away in this PR. Because still, some more testing needs to be done using these new workflows after this PR gets in.
Once we test this new way of calling Azure pipelines, after that it seems more promising to remove the existing workflows.

Mattjet27 (approved):
Changes look good!

PR #167, opened 2024-11-19
Fix Example Code Implementation in ConcurrentCodeExecution.md and MeasureVoltageMixed.cs

Mattjet27:

Fixes the example code in the ConcurrentCodeExecution.md user guide topic and similarly the implementation in the actual MeasureVoltageMixed.cs example code. Also, it adds a new `SimplerMeasureAndPublishCurrentSmuOrPpmuOrDmm` example method to the MeasureVoltageMixed.cs that highlights the new filterPins parameter added to the TSMSessionManager members, which was introduced in the 24.5.1 release. Note that this is intentionally added as a separate example to leave the existing `SimpleMeasureAndPublishCurrentSmuOrPpmuOrDmm` example as is for 24.5.0 users (although it was intentionally renamed to add OrDmm at the end of the method name since it includes a DMM measure in the implementation).

The current example code in the ConcurrentCodeExecution.md user guide topic and the similar implementation in the MeasureVoltageMixed.cs example code, has an incorrect

Manually tested the example code to ensure it is working as intended.

### DIFF
--- a/Examples/source/InstrumentAbstraction/MeasureVoltageMixed.cs
+++ b/Examples/source/InstrumentAbstraction/MeasureVoltageMixed.cs
@@ -9,7 +9,7 @@ namespace NationalInstruments.SemiconductorTestLibrary.Examples.InstrumentAbstra
 {
     /// <summary>
     /// This class contains examples of how to use the Instrument Abstraction extensions from the Semiconductor Test Library.
-    /// Specifically, how to measure current on pins that could be mapped to a DCPower instrument, Digital Pattern Instrument, or both.
+    /// Specifically, how to measure current on pins that could be mapped to a DCPower instrument, Digital Pattern Instrument, or Digital Multi Meter.
     /// Note that DCPower Instruments include both Source Measurement Units (SMUs) and Programmable Power Supplies (PPS) devices.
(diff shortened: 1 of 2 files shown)


lornazh:
@Mattjet27 When you say "fix implementation", there actually isn't an issue in original implementation, right? It seems the code change is to improve/simplify original implementation?

Mattjet27:
> @Mattjet27 When you say "fix implementation", there actually isn't an issue in original implementation, right? It seems the code change is to improve/simplify original implementation?

The code appears to fine does not need to be "fixed" from the compiler's point of view, but if you actually read it as a human, the variable naming is completely incorrect. The digital session bundle is being stored as 'smuPins' and the DCPower session bundle is being stored as 'ppmuPins'. As far as the compiled is concerned this works because in either case both objects get invoked with the same named member and arguments. However, this code is primarily used as an example for humans to reference and read through, not just compiled to be ran as code. It also has the potential to become copy and pasted. Therefore, I see this as "fixing" the code for the sake of it being used as an example, not just improving it or making the code more readable.

lornazh:
> the variable naming is completely incorrect. 

Ah that's the issue! I didn't notice that, it's very subtle. I thought you were just merging DMM Read and Publish calls to a ReadAndPublish call.

lornazh (approved):

PR #176, opened 2024-12-10
[WriterReviewNeeded] Update example code and user guide.

lornazh:

This pull request updates the example code and user guide page regarding sharing data between code modules.

We are updating these files so that they match the latest capabilities of the STL.

N/A


lornazh:
I'm not approving the pre pipeline for run because it will fail without Abstractions dll update. @Mattjet27

lornazh:
@Mattjet27 this change is ready for review.

Mattjet27:
> @Mattjet27 this change is ready for review.

@lornazh, this looks good, but I do have one small suggestion. I think it will be preferable and more readable if we did not use the `var` keyword in these examples, but instead used the explicit types we expect back from the Get methods `SiteData<double>` and `PinSiteData<double>`. I think for this case, it would be better practice.

Mattjet27 (changes requested):
> @Mattjet27 this change is ready for review.

@lornazh, this looks good, but I do have one small suggestion. I now think it will be preferable and more readable if we did not use the `var` keyword in these examples, but instead used the explicit types we expect back from the Get methods `SiteData<double>` and `PinSiteData<double>`. I think for this case, it would be better practice.

Mattjet27 (approved):

docs/UserGuide/advanced/SharingPinSiteDataBetweenCodeModules.md
```diff
@@ -39,16 +41,16 @@ public static void SecondCodeModule(
...
+    TSMSessionManager sessionManager = new TSMSessionManager(semiconductorModuleContext);
+    DigitalSessionsBundle digitalPins = sessionManager.Digital(pinName);
     digitalPins.BurstPattern(patternName);
-    SiteData<uint[]> measurement = digitalPins.FetchCaptureWaveform(waveformName, 1);
+    SiteData<uint[]> measurement = digitalPins.FetchCaptureWaveform(waveformName, samplesToRead);
 
-    var comparisonResults = measurement.Compare(ComparisonType.EqualTo, comparisonData);
-    semiconductorModuleContext.PublishResults(comparisonResults, "ComparisonResults");
+    SiteData<bool[]> comparisonResults = measurement.Compare<uint[], bool[]>(ComparisonType.EqualTo, comparisonData);
+    SiteData<bool> resultsToPublish = comparisonResults.Select(result => result.All(value => value == true));
```
lornazh, line 52: @Mattjet27 I modified the implementation of this method (fetch with the input `samplesToRead` parameter, and handling sequential compare so that publish won't fail). Please be aware and confirm this is what we want to demonstrate to customers.

Mattjet27: @lornazh, noted, this looks correct. However, my suggestion, as shown below, would be to provide a more meaningful variable name instead of `resultsToPublish`, and perhaps add a comment so that it is clear why this is being done. Also, I don't think you need to have `result.All(value => value == true)`, as this could be simplified to `result.All(value => value)`, unless it is intentional for added readability, 
```C#
// Publish the Aggregate Comparison Result, whether or not all samples in the comparison result are where found to be True.
SiteData<bool> aggregateComparisonResults = comparisonResults.Select(result => result.All(value => value));
```

Mattjet27 (approved):

docs/UserGuide/advanced/SharingPinSiteDataBetweenCodeModules.md
lornazh: Yes, using `value == true` is intentional for better readability. I simplified it as suggested, now that a comment is added.

docs/UserGuide/advanced/SharingPinSiteDataBetweenCodeModules.md
```diff
@@ -39,16 +41,17 @@ public static void SecondCodeModule(
...
     digitalPins.BurstPattern(patternName);
-    SiteData<uint[]> measurement = digitalPins.FetchCaptureWaveform(waveformName, 1);
+    SiteData<uint[]> measurement = digitalPins.FetchCaptureWaveform(waveformName, samplesToRead);
 
-    var comparisonResults = measurement.Compare(ComparisonType.EqualTo, comparisonData);
-    semiconductorModuleContext.PublishResults(comparisonResults, "ComparisonResults");
+    SiteData<bool[]> comparisonResults = measurement.Compare<uint[], bool[]>(ComparisonType.EqualTo, comparisonData);
+    // Publish the Aggregate Comparison Result: whether or not all samples in the comparison result are found to be True.
+    SiteData<bool> aggregateComparisonResult = comparisonResults.Select(result => result.All(value => value));
+    semiconductorModuleContext.PublishResults(resultsToPublish, "ComparisonResults");
```
Mattjet27, line 54: @lornazh, we missed this, it should have changed to "aggregateComparisonResult" to match the new variable name just above this. Please open a new PR to address this.

PR #187, opened 2025-01-06
[WriterReviewNeeded] User Story 2924579 Enable Source waveform - SetupNIDigitalPatternInst…

shrikantdeo-ni:
Justification
Currently, SetupNIDigitalPatternInstrument() method doesn't load any source waveform file to memory, because the method doesn't populate low level implemented flag to higher level.

Implementation
Add a boolean default parameter applySourceWaveformData, in the public static void SetupNIDigitalPatternInstrumentation() and pass it to InstrumentAbstraction.Digital.InitializeAndClose.Initialize()

In Teststand, go to the NI_SemiconductorTestStepTypes.ini. Modify the default module properties for the
NI_SemiconductorTest_SetupNIDigitalPatternInstrumentation by checking the checkbox for default value, as False.

Testing
1) Start Teststand application, In the Insertion Palette on left, locate the Step Types and select the option "dot net" from the drop down menu.
2) Expand the tree Semiconductor Module -> Common (dot net) -> Setup and Cleanup -> Setup NI Digital Instrumentation.
3) Right mouse button click on "Setup NI Digital Instrumentation" and from the popup menu, select "Insert Step".
4) In the layout "Steps: MainSequence", double click on the "Setup NI Digital Instrumentation".
5) In the layout "Step settings for Setup NI Digital Instrumentation" select the "Module" tab
5.1) Assembly should be browsed and selected to - NationalInstruments.SemiconductorTestLibrary.TestStandSteps.dll
5.2) In the dot Net invocation, check for the last parameter to be a boolean as below -
SetupNIDigitalPatternInstrumentation(NationalInstruments.TestStand.SemiconductorModule.CodeModuleAPI.ISemiconductorModuleContext, Boolean, System.String, System.String, Boolean) This definition should not show up in RED colour or be preceded by a RED coloured exclamation mark.
5.3) In the Digital Pattern Instrumentation table there is a list of parameters. Check for the parameter named "applySourceWaveformData". It should be a boolean type, with default value as False.


SemiconductorTestLibrary.TestStandSteps/source/SetupNIDigitalPatternInstrumentation.cs
```diff
@@ -18,15 +18,18 @@ public static partial class SetupAndCleanupSteps
         /// <param name="resetDevice">Whether to reset device during initialization.</param>
         /// <param name="levelsSheetToApply">The name of the levels sheet to apply.</param>
         /// <param name="timingSheetToApply">The name of the timing sheet to apply.</param>
+        /// <param name="applySourceWaveformData">The parameter to indicate propogation of source waveform data.</param>
```
lornazh, line 21: Suggest to put it: Whether to apply the data in waveform files to source waveforms.

lornazh (approved):

shrikantdeo-ni:
The parameter info in the header has been modified.

PR #192, opened 2025-01-09
[WriterReviewNeeded] Make VerifyTaskType public

lornazh:

Make the `VerifyTaskType` utility method public so that clients can make use of it.

Making the method public enables reusability.

N/A


Examples/source/InstrumentAbstraction/AnalogOutputDAQmx.cs
Mattjet27, line 1: @lornazh , I am thinking it would be better to follow our [recommended tip](https://ni.github.io/semi-test-library-dotnet/UserGuide/AddingUsingDirectives.html) for using the PreciseWait method here, which is to add the `using static NationalInstruments.SemiconductorTestLibrary.Common.Utilities;` directive at the top of the file, and reduces the statments on likes 47, 69, and 89 in this file to just `PreciseWait(timeInSeconds: 2)`. This avoids conflicts with any customer based Utilities classes or ones declared in namespaces being used such as in this case with the DAQmx Utilities class we just made public. It might be best to take this opportunity to make line 47 consistent with  69 and 89 in the file as well, such that is appears with the named parameter for readability: `PreciseWait(timeInSeconds: 2)`

Mattjet27 (changes requested):

Examples/source/InstrumentAbstraction/AnalogOutputDAQmx.cs
lornazh: Updated accordingly.

Mattjet27 (approved):

PR #199, opened 2025-01-26
UserStory 2982355 Remove dependency on Restricted Library

shrikantdeo-ni:
Eliminate the dependency on Restricted library by getting rid of SafeForEach() calls

To make the system safe at public access

1) Removed the "using NationalInstruments.Restricted;"
2) Replace the "SafeForEach" with the Do() method from Parallel - Common
3) Debug run each test method where the "SafeForEach()" was replaced

### Additional
Modified the exception.IfNotNull() because this one too depended on the Restricted library.


SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -716,14 +717,15 @@ void ConfigureOutputConnected()
                 sessionsBundle.ConfigureOutputConnected(values);
             }
 
-            var exception = Assert.Throws<AggregateException>(ConfigureOutputConnected);
-            exception.IfNotNull(x =>
```
lornazh, line 720: The `IfNotNull` check does not seem necessary, let's simply remove it.

lornazh (changes requested):
Now that you mentioned there are other usages of `NationalInstruments.Restricted` in these 2 test classes, I did a search and found more.
image
All of the extra test classes are using `IfNotNull` which I think could be simply removed. So let's fix all of them at once. Thanks.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -716,14 +717,15 @@ void ConfigureOutputConnected()
...
+            AggregateException excAggregate = null;
+            excAggregate = Assert.Throws<AggregateException>(ConfigureOutputConnected);
+            if (null != excAggregate.InnerException)
```
lornazh, line 722: This is changing the test logic.

The test is verifying whether an aggregate exception that wraps a certain inner exception is thrown. With this change, the test passes when the criteria is not meet.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PPMUTests.cs
```diff
@@ -668,16 +667,14 @@ public void SessionsInitialized_MeasureAndPublishCurrent_ThrowsExecptionAndRetur
 
             void MeasureAndPublishCurrent() => sessionsBundle.MeasureAndPublishCurrent("CurrentMeasurments");
 
-            var exception = Assert.Throws<AggregateException>(MeasureAndPublishCurrent);
+            AggregateException excAggregate = Assert.Throws<AggregateException>(MeasureAndPublishCurrent);
+
```
lornazh, line 671: Please revert the empty line 671.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
shrikantdeo-ni: Removed the check

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DAQmx/AnalogOutputFunctionGenerationTests.cs
```diff
@@ -43,14 +42,11 @@ public void UnsupportedDevice_GenerateSineWaveThrowsException()
             var sessionManager = Initialize("DAQmxTests.pinmap");
             var tasksBundle = sessionManager.DAQmx("VDD");
 
-            var exception = Assert.Throws<AggregateException>(() => GenerateSineWave(tasksBundle));
-            exception.IfNotNull(x =>
+            AggregateException excAggregate = Assert.Throws<AggregateException>(() => GenerateSineWave(tasksBundle));
```
lornazh, line 45: I don't think it's necessary to make the exception type explicit (line 45 and line 46).
Also, "exc" does not feel like a good abbreviation for "exception". If you do want to make it explicit, I would recommend "aggregateException".

lornazh (approved):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
shrikantdeo-ni: Removed the check

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PPMUTests.cs
shrikantdeo-ni: Done

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DAQmx/AnalogOutputFunctionGenerationTests.cs
shrikantdeo-ni: Changed to "aggregateException"

PR #201, opened 2025-01-30
Remove Dependency on Restricted Namespace from Examples

Mattjet27:

Removes the dependency on the `NationalInstruments.Restricted` namespace by updating the `GetDeviceInfo` private method implementation in the `PatternTriggeredDmmMeasurement` example to use the native `System.Array.IndexOf` method.

Ensure that public code is not using the `NationalInstruments.Restricted` namespace.

Verified that all code builds without error. No additional testing was done as the `System.Array.IndexOf` method is functionally equivalent to the `NationalInstruments.Restricted.EnumerableExtensions.IndexOf`.


lornazh (commented):
Reviewed.

PR #206, opened 2025-02-12
Add STL-Specific Code Analyzer Package For Documenting Class Constructors

Shalini-Subramanian:
- Integrated the analyzer through the NuGet package.
- Added packages folder to repository and included NuGet package.
- Added nuget.config to fetch the NuGet package from within the solution files.
- Added package reference to each CSProject file to reference the NuGet package.

Whenever a class does not contain at least one constructor, its documentation will not be visible to the user when an instance is created. Currently, this is validated manually. In this PR, I have addressed this issue by using a NuGet package with an analyzer created in the previous to detect missing class constructors.

- Error popped up and added the required fixes to address the missing of the constructor.


.gitignore
```diff
@@ -7,7 +7,6 @@
 /SemiconductorTestLibrary.TestStandSteps/source/net48
 
 # Nuget packages & config
-/**/packages
```
Shalini-Subramanian, line 10: Since the packages folder has to be included to the repo, removed the ignore command. @lornazh Please let me know if this is fine

lornazh: I suggest we added it back when your change goes in.

nuget.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageRestore>
+    <add key="enabled" value="True" />
+    <add key="automatic" value="True" />
+  </packageRestore>
+  <packageSources>
+    <add key="NuGet" value="https://www.nuget.org/api/v2/" />
+    <add key="STLNuGet" value=".\packages" />
```
Mattjet27, line 9: I think this should just be "packages"

Mattjet27, line 9: I suggest changing "STLNuGet" to "withinMyProject", "SolutionFiles", otherwise I feel it implies that there might be a different online server required for this.

packages/SemiconductorTestLibrary.Analyzers.1.0.0.nupkg
Mattjet27, line 1: I feel this should be name spaced as `NationalInstruments.SemiconductorTestLibrary.Analyzers.1.0.0`

.gitignore
Mattjet27: FYI @Shalini-Subramanian, @lornazh, you should not need to have to add it back. You could have simply used the `git add --force packages/` command (from the repo root directory). This would have left this file untouched, but allowed the specific package we wanted in the repo to now be tracked by git. 

https://stackoverflow.com/questions/8006393/force-add-despite-the-gitignore-file

Mattjet27 (changes requested):

.gitignore
Shalini-Subramanian: Updated accordingly

packages/SemiconductorTestLibrary.Analyzers.1.0.0.nupkg
Shalini-Subramanian: Updated

nuget.config
Shalini-Subramanian: Updated

nuget.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageRestore>
+    <add key="enabled" value="True" />
+    <add key="automatic" value="True" />
+  </packageRestore>
+  <packageSources>
+    <add key="NuGet" value="https://www.nuget.org/api/v2/" />
+    <add key="STLNuGetInSolutionFiles" value="packages" />
```
Mattjet27, line 9: @Shalini-Subramanian, I am not sure we need the STL part, it could just be, "NuGetInSolutionFiles".

Mattjet27 (approved):

nuget.config
Shalini-Subramanian: Updated

lornazh (approved):

PR #234, opened 2025-04-05
Update auto-comment-closed-wi.yml Workflow: Tag Filtering and Issue State Validation

Utkarsh-NI-Emerson:

- Processes only work items with the "Comment on GitHub" tag.
- Retrieves the current state of the linked GitHub issue before commenting.
- Skips posting comments on GitHub issues that are already closed.
- Refactors the commenting step to use asynchronous functions.

- Ensures only intended work items are processed.
- Prevents redundant comments on closed issues.

- Executed the updated workflow on a sample set of work items.
- Verified that only work items with the "Comment on GitHub" tag are processed.
- Confirmed that closed GitHub issues are correctly skipped.
- Verified logs to ensure expected execution.

### DIFF
--- a/.github/workflows/auto-comment-closed-wi.yml
+++ b/.github/workflows/auto-comment-closed-wi.yml
@@ -54,7 +54,11 @@ jobs:
               -X GET "https://dev.azure.com/$AZDO_ORG/$AZDO_PROJECT/_apis/wit/workitems/$id?\$expand=relations&api-version=7.1-preview.2")
             EXISTING_TAGS=$(echo "$WI_DETAILS" | jq -r '.fields["System.Tags"]')
             if echo "$EXISTING_TAGS" | grep -q "GitHubCommented"; then
-              echo "Work Item $id already has GitHubCommented tag. Skipping."
+              echo "Work Item $id already has 'GitHubCommented' tag. Skipping."
+              continue
+            fi
+            if ! echo "$EXISTING_TAGS" | grep -q "Comment on GitHub"; then
+              echo "Work Item $id does not have 'Comment on GitHub' tag. Skipping."
               continue
             fi
             ISSUE_URL=$(echo "$WI_DETAILS" | jq -r '.relations[]? | select(.rel=="ArtifactLink" and .attributes.name=="GitHub Issue") | .url')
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-comment-closed-wi.yml
```diff
@@ -54,7 +54,11 @@ jobs:
               -X GET "https://dev.azure.com/$AZDO_ORG/$AZDO_PROJECT/_apis/wit/workitems/$id?\$expand=relations&api-version=7.1-preview.2")
             EXISTING_TAGS=$(echo "$WI_DETAILS" | jq -r '.fields["System.Tags"]')
             if echo "$EXISTING_TAGS" | grep -q "GitHubCommented"; then
-              echo "Work Item $id already has GitHubCommented tag. Skipping."
+              echo "Work Item $id already has 'GitHubCommented' tag. Skipping."
+              continue
+            fi
```
Mattjet27, line 59: Is there any particular reason for the tags being differently formatted (one with spaces, one with out?): 'Comment on GitHub' vs. 'GitHubCommented' ?

Mattjet27 (approved):

.github/workflows/auto-comment-closed-wi.yml
Utkarsh-NI-Emerson: @Mattjet27, we wanted to keep both the tags as distinguished as possible, apart from that there is no other reason for the difference in formatting.
We can change it if this feels incorrect.

Thanks.

Mattjet27: @Utkarsh-NI-Emerson, can you please clarify what the differences are between these tags then, so we can have that documented here? It is not clear to me.

Utkarsh-NI-Emerson: @Mattjet27 We are using "GitHubCommented' tag just for a redundancy check so that the workflow does not comment on the same Issue twice.
While "Comment on GitHub" tag is for developers to know that a comment would be made for a particular work item, if they don't want to send out a comment on issue for the closure of this work item then they HAVE to remove this tag. Additionally, this tag is checked while making a comment on GitHub Issue. So, this tag provides a functionality as well as a check.

This was our reason behind these tags, but I can move forward with your suggestion as well if you confirm so.

Thanks.

Mattjet27: @Utkarsh-NI-Emerson, okay so it sounds like the internal items would have the `Comment on GitHub` it ensure it gets commented on, unless the developer manually removes it before closing the item. However, if the tag exists when closing the item, the comment will be made, and subsequently the `GitHubCommented` tag will be added, which will effectively mean that when closed out, there will be two tags, both `Comment on GitHub` and `GitHubCommented`. Do I understand that correct, or am I missing something? Assuming I understand correctly, then my suggestion is for the tags to be `CommentOnGitHub` and `GitHubCommentWasMade`. 

That said, I feel it would better if we just had the single `CommentOnGitHub` tag, and instead of adding the second tag (which feels quite redundant), we simply add a comment to the AzDO item after it has been closed to say that a GitHub comment was made, and ideally links to that comment. That way, we ensure if something goes wrong, there is easy traceability for the developer to correct and delete the comment if needed.

Utkarsh-NI-Emerson: Yes @Mattjet27, your understanding is correct. I will go ahead and change the `Comment on GitHub` tag to `CommentOnGitHub` and `GitHubCommented` tag to `GitHubCommentWasMade` respectively.

For the second part, the `GitHubCommented` tag is added by the workflow itself, while making the comment, and is used as a check during the next run. This is so that there are no redudant comments on the same Issue.
This is a condition check we have used, as the workflow fetches the closed work items in last 2 days and it is scheduled to run every 24 hours. Just to be on the same page the `GitHubCommented` tag is NOT intended to notify the developers about the comment being made. Also, for a quicker access we are already including the Issue link in the description of the Work Item.

Thanks.

Utkarsh-NI-Emerson: Hey @Mattjet27, I have changed the `Comment on GitHub` tag to `CommentOnGitHub` and `GitHubCommented` tag to `GitHubCommentWasMade` respectively.
Thanks.

PR #276, opened 2025-06-16
Shared Pin | Update Supporting Materials for Custom Instruments

deepika-ni:

This PR adds the supporting materials for testing the `Shared Pin` support of **Custom Instruments**, specifically including the pin map configurations required to simulate hardware behavior and validate Shared Pin functionality.

Added a `.pinmap` file with Shared Pin configuration under TestAssests/Supporting Materials folder.

N/A

### DIFF
--- a/TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
+++ b/TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
@@ -1,0 +1,147 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PinMap xmlns="http://www.ni.com/TestStand/SemiconductorModule/PinMap.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" schemaVersion="1.2">
+	<Instruments>
+		<Instrument name="SMU_4147_C1_S11" instrumentTypeId="SessionPerChannelInst">
+			<Channel id="ch0" />
+			<Channel id="ch1" />
+			<Channel id="ch2" />
+			<Channel id="ch3" />
+			<Channel id="ch4" />
+			<Channel id="ch5" />
+			<Channel id="ch6" />
+			<Channel id="ch7" />
+		</Instrument>
+		<Instrument name="SMU_4147_C2_S10" instrumentTypeId="SessionPerChannelInst">
+			<Channel id="ch0" />
+			<Channel id="ch1" />
+			<Channel id="ch2" />
+			<Channel id="ch3" />
+			<Channel id="ch4" />
+			<Channel id="ch5" />
+			<Channel id="ch6" />
+			<Channel id="ch7" />
+		</Instrument>
+		<Instrument name="SMU_4147_C3_S12" instrumentTypeId="SessionPerInstrument">
+			<ChannelGroup id="allChannels">
+				<Channel id="0" />
+				<Channel id="1" />
+				<Channel id="2" />
+				<Channel id="3" />
+				<Channel id="4" />
+				<Channel id="5" />
+				<Channel id="6" />
+				<Channel id="7" />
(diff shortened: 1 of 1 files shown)


TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
```diff
@@ -0,0 +1,129 @@
...
+		<Instrument name="SMU_4147_C3_S12" instrumentTypeId="SessionPerInstrument">
+			<ChannelGroup id="allChannels">
+				<Channel id="0" />
+				<Channel id="1" />
+				<Channel id="2" />
+				<Channel id="3" />
+				<Channel id="4" />
+				<Channel id="5" />
+				<Channel id="6" />
+				<Channel id="7" />
```
lornazh, line 33: Is it intentional to directly use a number instead of using a leading "ch" prefix?

deepika-ni: Added similar to this `Session per Instrument.pinmap` file but modified now to maintain consistency.

TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="utf-8"?>
+<PinMap xmlns="http://www.ni.com/TestStand/SemiconductorModule/PinMap.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" schemaVersion="1.2">
+	<Instruments>
+		<Instrument name="SMU_4147_C1_S11" instrumentTypeId="SessionPerChannelInst">
```
ni-ramya, line 4: SMU is not a custom instrument. So let's not name it like this. Let's name it generically like 'CUSTOM1_C1_S11'

ni-ramya (changes requested):

TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
```diff
@@ -0,0 +1,110 @@
...
+			<ChannelGroup id="Grp2">
+				<Channel id="ch4" />
+				<Channel id="ch5" />
+				<Channel id="ch6" />
+				<Channel id="ch7" />
+			</ChannelGroup>
+		</Instrument>
+	</Instruments>
+	<Pins>
+		<DUTPin name="A" />
```
Shalini-Subramanian, line 78: Can we have a proper pin name 
```suggestion
		<DUTPin name="PinA" />
```

TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
lornazh: @ni-ramya Thought about that, I think the reason is, instrument name must be something that exists in offline config, otherwise the initialization fails.

@deepika-ni Correct me if this is not right.

lornazh (approved):

TestAssets/Supporting Materials/Pin Maps/CustomInstrumentsSharedPin.pinmap
deepika-ni: @lornazh  Yes, Lorna, you're right. We either need to update the offline config or keep the current instrument name as it is. I recommend we stick with the existing name, @ni-ramya  what are your thoughts on that?

deepika-ni: I prefer to stick with existing instrument name, thus avoiding the changes in offline config file.

PR #285, opened 2025-06-25
Shared Pin | Exclude Shadow Sites 

deepika-ni:

This PR contains the change where the shadow sites of shared channel will be excluded from being processed. As they have same channel as primary site, doing the same operation again is unnecessary.

This PR eliminates the redundant operation being performed on same channel with same values.

Tested with existing test cases.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -628,7 +628,7 @@ private static void AbortAndConfigure(this DCPowerSessionInformation sessionInfo
             }
             else
             {
-                foreach (var sitePinInfo in sessionInfo.AssociatedSitePinList)
+                foreach (var sitePinInfo in sessionInfo.AssociatedSitePinList.GroupBy(sitePin => sitePin.IndividualChannelString).Select(group => group.First()))
```
lornazh, line 631: Couldn't we simply wrap the logic inside a `sitePinInfo.SkipOperations` check? I believe that's the exact purpose of introducing `SkipOperations`.

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
deepika-ni: Yeah, that makes sense, updated the query to make use of `SkipOperations`.

Initially, I considered the possibility that `SkipOperations` might not always be updated, example in scenarios where the primary site is filtered out and the `sitepinlist` is used directly. However, upon reviewing the logic, I noticed that whenever a filter is applied, session is recreated, which ensures that `ChannelSharedInfo` is updated as part of the filtering process.

Are there any other scenarios you can think of where `ChannelSharedInfo` might not be updated? If so, please let me know.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DAQmx/Configure.cs
```diff
@@ -130,7 +130,7 @@ public static PinSiteData<double> GetSampleClockRate(this DAQmxTasksBundle tasks
         {
             return tasksBundle.DoAndReturnPerSitePerPinResults(taskInfo =>
             {
-                return Enumerable.Repeat(taskInfo.Task.Timing.SampleClockRate, taskInfo.AssociatedSitePinList.Count).ToArray();
+                return Enumerable.Repeat(taskInfo.Task.Timing.SampleClockRate, taskInfo.AssociatedSitePinList.Where(sitePin => !sitePin.SkipOperations).Count()).ToArray();
```
Shalini-Subramanian, line 133: In this case, we're using Count, right? As I understand, Count returns the number of primary and non-shared channels count.
Now, let’s say site 0 and site 1 are shared, and site 2 is non-shared. In this scenario, won't `Enumerable.Repeat` be applied to the first two sites based on the count (i.e., site 0 and site 1)? Ideally, it should be applied to site 0 and site 2 right?

Could you please confirm if my understanding is correct, or if I’m missing something?

lornazh: @Shalini-Subramanian Your understanding is partially correct.

The body of the lamda function that `DoAndReturnPerSitePerPinResults` calls, is responsible for returning the results for the shared primary and non-shared channels. So it's correct to return 2 values in the case you mentioned, but the values are for site 0 and site 2 (i.e. site number has nothing to do with the array index). Which sites to map these values to, is determined by the `ParallelExecution.FillResultsDictionary` method, in which [the `ChannelSharedInfo` is examined](https://dev.azure.com/ni/DevCentral/_git/srd-sebu-mixed-signal-apt?path=/SemiconductorTestLibraryDotNET/SemiconductorTestLibraryAbstractions/source/Common/ParallelExecution.cs&version=GBmain&line=490&lineEnd=497&lineStartColumn=13&lineEndColumn=14&lineStyle=plain&_a=contents) to fill in values for shared shadow channel.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -536,7 +536,7 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
             var session = sessionInfo.Session;
             var lockObject = new object();
 
-            int channelCount = sessionInfo.AssociatedSitePinList.Count;
+            int channelCount = sessionInfo.AssociatedSitePinList.Select(sitePin => !sitePin.SkipOperations).Count();
```
lornazh, line 539: Should this be `Where`?

deepika-ni: As we just needed the count I used `Select`, which now updated to `Where` to maintain consistency.

Shalini-Subramanian (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
deepika-ni: Sure Lorna, got it.

Issue #292, opened 2025-07-04
[BUG] DCPowerSessionsBundle ClearTriggers extension method generates exception on PXIe-4147


ktvanzwol:
**Describe the bug**
Calling ClearTriggers on a DCPowerSessionBundle with a PXIe-4147 results in thisexception:
```
System.AggregateException
  HResult=0x80131500
  Message=One or more errors occurred.
  Source=mscorlib
  StackTrace:
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Parallel.ForWorker[TLocal](Int32 fromInclusive, Int32 toExclusive, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Func`4 bodyWithLocal, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEachWorker[TSource,TLocal](IEnumerable`1 source, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEach[TSource](IEnumerable`1 source, Action`1 body)
   at NationalInstruments.SemiconductorTestLibrary.Common.ParallelExecution.Do[TSessionInformation](ISessionsBundle`1 sessionsBundle, Action`2 action)
   at XFDefaultTemplate.TestSteps.UTCPowerUp(ISemiconductorModuleContext tsmContext, String[] dcPins, String digitalPins, String triggerTerminal, String eventCode) in C:\Users\Public\Documents\National Instruments\TestStand 2023 (64-bit)\Components\Tools\STS Project Creation Tool\Templates\X-FAB UTC v4.0 Template\Code Modules\TestSteps\TemplateTestStep.cs:line 54

Inner Exception 1:
AggregateException: One or more errors occurred.

Inner Exception 2:
OperationNotSupportedException: ModularInstruments.NIDCPower: ModularInstruments.NIDCPower: IVI: (Hex 0xBFFA000C) Attribute ID not recognized.

Device: SMU_P142_CH0-3
Error code: -1074135028

Inner Exception 3:
[...]

PR #301, opened 2025-07-17
Bug Fix: DCPowerSessionsBundle ClearTriggers extension method generates exception on PXIe-4147

Tanish-NI:
- Added a read-only Dictionary in Utilities.cs to be referred to as a map to unsupported SMU types for each type of trigger. Reference NI documentation: https://www.ni.com/docs/en-US/bundle/ni-dcpower-labview-api-ref/page/instr-lib/nidcpower/nidcpower-rc/nidcpower/supported-properties.html
- In the current code, there is no ShutDown trigger support, but the dictionary is fulfilled with that as well to support future development.
- Used the dictionary and exposed 1 public APIs - GetUnsupportedTriggerTypes, in TriggerAndEvents.cs to handle exceptions caused by unsupported models in configure or clear triggers.
-  Refactored TriggerAndEvents.cs to simplify the unsupported models check and removed some public APIs.

To fix the GitHub issue raised - https://github.com/ni/semi-test-library-dotnet/issues/292.

- Added unit tests to make sure any of the unsupported SMU types don't throw an exception when clear triggers. 
- To support unit tests creation, updated the MixedSignal offlineconfig and created a new pinmap MultipleSMUTypes.pinmap.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -488,8 +488,7 @@ public static void ConfigureSourceTriggerDisable(this DCPowerSessionInformation
             sessionInfo.DoForSupportedModels(
                 channelString,
                 modelString,
-                // Might want to move this out of this method and into to a readonly dictionary of some kind in the future.
-                new string[] { DCPowerModelStrings.PXI_4110, DCPowerModelStrings.PXI_4130, DCPowerModelStrings.PXIe_4154 },
```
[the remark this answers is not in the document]
lornazh: Could be a mistake we made earlier.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -255,8 +265,7 @@ public static void ConfigureMeasureTriggerDigitalEdge(this DCPowerSessionInforma
             sessionInfo.DoForSupportedModels(
                 channelString,
                 modelString,
-                // Might want to move this out of this method and into to a readonly dictionary of some kind in the future.
-                new string[] { }, // Placeholder for now, use DCPowerModelStrings
+                GetUnSupportedModelStrings(TriggerType.MeasureTrigger),
```
lornazh, line 268: Now that we created a dedicated method, we could make `DoForSupportedModels` take in just the trigger type, and invoke `GetUnSupportedModelStrings` inside `DoForSupportedModels` for simplicity.

Tanish-NI: Alright... In that case I think we don't even need the API `GetUnSupportedModelStrings`. Just this `IsTriggerTypeSupported` will be sufficient as we have specific modelString and triggerType information inside `DoForSupportedModels`, right?

lornazh: Yes I think that makes sense.

Tanish-NI: Updated `DoForSupportedModels` to take triggerType as input and removed `GetUnSupportedModelStrings` API.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -214,12 +214,21 @@ public static void ClearTriggers(this DCPowerSessionsBundle sessionsBundle)
         {
             sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
             {
-                if (pinSiteInfo.ModelString != DCPowerModelStrings.PXI_4110)
+                sessionInfo.AllChannelsOutput.Control.Abort();
```
lornazh, line 217: Theoretically the logic is changed.

We originally only abort when triggers are to be cleared. - @Mattjet27 is that some logic we want to maintain?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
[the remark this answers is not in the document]
lornazh: @Mattjet27 Please be aware that we're correcting the existing logic according to NI documentation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -1,12 +1,91 @@
...
             return string.Join(",", channelString.Split(',').Where(s => !s.Contains($"/{channelToExclude}")));
         }
+
+        public static bool IsTriggerTypeSupported(string modelString, TriggerType triggerType)
+        {
+            if (_triggerTypeToUnsupportedModelStringMap.ContainsKey(triggerType))
+            {
+                return !_triggerTypeToUnsupportedModelStringMap[triggerType].Contains(modelString);
+            }
+            return false;
```
lornazh, line 88: Does it make more sense to return `true` or `false` when trigger type is not specified in the dictionary?

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
Tanish-NI: First of all, there is very less chance that the trigger type is not present in dictionary because we included each trigger type already, I added this just to be on safer side.
Second, I made it to return false, as if the trigger type is not present in dictionary, then we are not sure if it's supported or not. Hence to avoid exception in worst case, I believe returning false is more reliable... As we are considering the dictionary as single source of truth.

lornazh: What I was thinking is, since the dictionary is the source of truth of unsupported models, it might be better if we return true when no trigger types or model strings can be found.

One extreme example is, one day when a new type of trigger type is added on driver side, do we want it the driver code to be called or silently not called by default?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -1,12 +1,91 @@
-﻿using System.Linq;
+﻿using System.Collections.Generic;
+using System.Linq;
 
 namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCPower
 {
     internal static class Utilities
     {
+        private static readonly Dictionary<TriggerType, IList<string>> _triggerTypeToUnsupportedModelStringMap = new Dictionary<TriggerType, IList<string>>
```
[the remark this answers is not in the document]
lornazh: @Mattjet27 @Tanish-NI I'd like to add one more point.

Let's say one day, driver adds a new trigger type or a new SMU model.

If STL keeps a white list, then without modifying the list in STL, no operation related to this new trigger type or model can get called. Which means STL needs to be patched immediately to support driver upgrade.

If STL keeps a black list, then without modifying the list in STL, the supported operations related to this new trigger type or model are allowed to be called, although there could be errors when unsupported operations are called. Customers can work around those errors by adding checks in their code to avoid calling STL extension methods for this new trigger type or model.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
Tanish-NI: Thanks for the confirmation @Mattjet27. To handle this request efficiently, I exposed a new API `GetSupportedTriggerTypes` to get all supported trigger types for a model. This way I don't need to call the existing `IsTriggerTypeSupported` API multiple times to check if there is at least 1 trigger type supported for the model string to call abort.
cc: @lornazh

TestAssets/Supporting Materials/Pin Maps/MultipleSMUDevices.pinmap
```diff
@@ -0,0 +1,82 @@
...
+			<ChannelGroup name="DCPowerChannelGroup_4147" channels="0, 1, 2, 3" />
+		</NIDCPowerInstrument>
+		<NIDCPowerInstrument name="SMU_4154_C3_S04" numberOfChannels="2">
+			<ChannelGroup name="DCPowerChannelGroup_4154" channels="0, 1" />
+		</NIDCPowerInstrument>
+		<NIDCPowerInstrument name="SMU_4162_C5_S10" numberOfChannels="12">
+			<ChannelGroup name="DCPowerChannelGroup_4162" channels="0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11" />
+		</NIDCPowerInstrument>
+		<NIDCPowerInstrument name="SMU_4163_C5_S13" numberOfChannels="24">
+			<ChannelGroup name="DCPowerChannelGroup_4163" channels="0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23" />
```
lornazh, line 45: I think we could simplify with not explicitly enumerate channels.
image
image

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
lornazh: @Tanish-NI I was actually thinking about duplicating the `Abort` call inside every if trigger type supported block, which is more straight forward and readable :)

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -1,12 +1,105 @@
...
+        {
+            if (_triggerTypeToUnsupportedModelStringMap.ContainsKey(triggerType))
+            {
+                return !_triggerTypeToUnsupportedModelStringMap[triggerType].Contains(modelString);
+            }
+            // If something is not present in unsupported model strings list, then assume its supported.
+            return true;
+        }
+
+        public static IList<TriggerType> GetSupportedTriggerTypes(string modelString)
```
lornazh, line 92: This method by nature has a problem, which is, trigger types that all models support won't be returned (newly added trigger types won't be returned either unless we update the unsupported list).

So I'm a little hesitate to add this helper method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
Tanish-NI: But the reason I didn't go with that approach is, I am thinking if a model supports more than 1 trigger type, then `Abort` will be called multiple times, which is not required, right?

lornazh: No it's not required and it's a good point.

Tanish-NI: Alternatively, I can propose that we hardcode the 4110 and 4130 to not call about the abort; the rest all call abort. However, I didn't adopt this approach because I want the dictionary to be the single source of truth.

One more way I thought of is to expose an API `AnyTriggerTypeSupported`; this will just return true if at least one trigger type is supported by iterating the dictionary. Its is also a good way.... but then we will query the dictionary multiple times in the `ClearTrigger` method :)

Tanish-NI: What will the condition for abort look like? We need to check if the returned dictionary contains all the trigger types, which will again cause problems when a new trigger type is added.

Does abort need to be called before we set trigger type to None? If not, then we can use a variable to check if at least one if condition is executed, then call abort. Like this?

```
public static void ClearTriggers(this DCPowerSessionsBundle sessionsBundle)
        {
            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
            {
                bool anyTriggerTypeSupported = false;
                if (IsTriggerTypeSupported(pinSiteInfo.ModelString, TriggerType.PulseTrigger))
                {
                    anyTriggerTypeSupported  = true;
                    sessionInfo.AllChannelsOutput.Triggers.PulseTrigger.Type = DCPowerPulseTriggerType.None;
                }
                if (IsTriggerTypeSupported(pinSiteInfo.ModelString, TriggerType.SequenceAdvanceTrigger))
                {
                    anyTriggerTypeSupported  = true;
                    sessionInfo.AllChannelsOutput.Triggers.SequenceAdvanceTrigger.Type = DCPowerSequenceAdvanceTriggerType.None;
                }
                if (IsTriggerTypeSupported(pinSiteInfo.ModelString, TriggerType.SourceTrigger))
                {
                    anyTriggerTypeSupported  = true;
                    sessionInfo.AllChannelsOutput.Triggers.SourceTrigger.Type = DCPowerSourceTriggerType.None;
                }
                if (IsTriggerTypeSupported(pinSiteInfo.ModelString, TriggerType.StartTrigger))
                {
                    anyTriggerTypeSupported  = true;
                    sessionInfo.AllChannelsOutput.Triggers.StartTrigger.Type = DCPowerStartTriggerType.None;
                }
                if(anyTriggerTypeSupported)
                {
                sessionInfo.AllChannelsOutput.Control.Abort();
                }
            });
        }
```

Mattjet27: @Tanish-NI,  to confirm, the `Abort` operation needs o be called before setting the `TriggerType` property object.

lornazh: > What will the condition for abort look like? We need to check if the returned dictionary contains all the trigger types, which will again cause problems when a new trigger type is added.

@Tanish-NI This is what I'm thinking about
```
sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
{
    var triggerTypeUnsupported = GetUnsupportedTriggerTypes(pinSiteInfo.ModelString);
    var triggerTypesToDealWith = new List<TriggerType>() { TriggerType.PulseTrigger, TriggerType.SequenceAdvanceTrigger, TriggerType.SourceTrigger, TriggerType.StartTrigger };
    if (triggerTypesToDealWith.Except(triggerTypeUnsupported).Any())
    {
        sessionInfo.AllChannelsOutput.Control.Abort();
        if (!triggerTypeUnsupported.Contains(TriggerType.PulseTrigger))
        {
            sessionInfo.AllChannelsOutput.Triggers.PulseTrigger.Type = DCPowerPulseTriggerType.None;
        }
        if (!triggerTypeUnsupported.Contains(TriggerType.SequenceAdvanceTrigger))
        {
            sessionInfo.AllChannelsOutput.Triggers.SequenceAdvanceTrigger.Type = DCPowerSequenceAdvanceTriggerType.None;
        }
        if (!triggerTypeUnsupported.Contains(TriggerType.SourceTrigger))
        {
            sessionInfo.AllChannelsOutput.Triggers.SourceTrigger.Type = DCPowerSourceTriggerType.None;
        }
        if (!triggerTypeUnsupported.Contains(TriggerType.StartTrigger))
        {
            sessionInfo.AllChannelsOutput.Triggers.StartTrigger.Type = DCPowerStartTriggerType.None;
        }
    }
});
```
With this implementation:

- If new trigger type is added, and we want to deal with it, then we need to modify the logic of this method anyway (to set the new trigger type to None).
- If new trigger type is added, and we don't want to deal with it yet, then the logic does not need to be modified.
- If new model is added and it's supported, then the logic could still work without modifications.
- If new model is added and it's not supported, then we need to modify the unsupported dictionary but the logic of this method could keep the same.

Does that make sense?

lornazh:
@Tanish-NI It seems we need to apply same logic to `SendSoftwareEdgeTrigger` and `WaitForEvent` no matter what that logic ends up look like, right?

And for all of other sessions bundle extension methods with `Abort` call, that `Abort` happens as always even when no driver method is invoked afterwards, which we want to avoid? @Mattjet27

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
Tanish-NI: Ok, makes sense. I agree to this approach now. Thanks, @lornazh!

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -214,12 +215,25 @@ public static void ClearTriggers(this DCPowerSessionsBundle sessionsBundle)
         {
             sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
             {
-                if (pinSiteInfo.ModelString != DCPowerModelStrings.PXI_4110)
+                IList<TriggerType> triggerTypesSupported = GetSupportedTriggerTypes(pinSiteInfo.ModelString);
+                if (triggerTypesSupported.Count > 0)
                 {
                     sessionInfo.AllChannelsOutput.Control.Abort();
```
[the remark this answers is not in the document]
Tanish-NI: @lornazh, so you are saying that ClearTrigger should be updated to look like this?
image

lornazh:
> > @Tanish-NI It seems we need to apply same logic to `SendSoftwareEdgeTrigger` and `WaitForEvent` no matter what that logic ends up look like, right?
> 
> @lornazh I didn't apply the conditional logic to `SendSoftwareEdgeTrigger` because here the user himself is passing the TriggerType. Hence, if the model does not support a trigger type, then the exception at the user end should be expected.

User is passing `TriggerType` to `ConfigureTriggerSoftwareEdge` as well, and if the model does not support a trigger type, we don't error currently. So I think we should be consistent.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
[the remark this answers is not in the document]
lornazh: Yes that's what I'm expecting.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -33,28 +35,28 @@ public static void ExportSignal(this DCPowerSessionsBundle sessionsBundle, DCPow
         /// <param name="triggerType">The type of the trigger.</param>
         public static void SendSoftwareEdgeTrigger(this DCPowerSessionsBundle sessionsBundle, TriggerType triggerType)
```
Tanish-NI, line 36: @lornazh I have updated the `SendSoftwareEdgeTrigger` to also consider unsupported conditional check, please have a look. If you want for Events as well the that can be considered for separate PR or work item.
I have one question.... I see there is no unit tests for `SendSoftwareEdgeTrigger` currently, is that intentional? I tried creating unit tests for this, but some of Triggers like `StartTrigger` is failing with Error - `Start Trigger Type must be set to None(disabled) if Source Mode attribute/ property is set to Single Point.`. I talked with Anish @nitamil as well, it seems the Seq mode is not yet supported in STL? So, should I skip unit test creation for `SendSoftwareEdgeTrigger` for now?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -214,13 +216,27 @@ public static void ClearTriggers(this DCPowerSessionsBundle sessionsBundle)
         {
             sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
             {
-                if (pinSiteInfo.ModelString != DCPowerModelStrings.PXI_4110)
+                var triggerTypesUnsupported = GetUnsupportedTriggerTypes(pinSiteInfo.ModelString);
+                var triggerTypesToDealWith = new List<TriggerType>() { TriggerType.PulseTrigger, TriggerType.SequenceAdvanceTrigger, TriggerType.SourceTrigger, TriggerType.StartTrigger };
```
Mattjet27, line 220: I suggest renaming this variable to `triggerTypesToClear`, this is more informative that "DealWith" in my opinion.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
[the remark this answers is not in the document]
Mattjet27: @Tanish-NI and @lornazh, thank you for elaborating, that all makes sense to me. Please proceed as planned.

Mattjet27 (approved):
@Tanish-NI, I only have one suggestion, otherwise I approved. Just pending @lornazh's final approval. Thanks!

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -1,12 +1,103 @@
...
             return string.Join(",", channelString.Split(',').Where(s => !s.Contains($"/{channelToExclude}")));
         }
+
+        public static bool IsTriggerTypeSupported(string modelString, TriggerType triggerType)
+        {
+            if (_triggerTypeToUnsupportedModelStringMap.ContainsKey(triggerType))
+            {
+                return !_triggerTypeToUnsupportedModelStringMap[triggerType].Contains(modelString);
+            }
+            // If something is not present in unsupported model strings list, then assume its supported.
```
lornazh, line 88: "then assume **it's** supported".

lornazh (changes requested):
I mentioned in another comment that we'd better apply the same logic we put in `ClearTriggers` extension method so that we only abort when there are supported trigger types.

I think we should check all `Abort` calls.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
lornazh: Sorry I don't remember why there isn't unit tests for `SendSoftwareEdgeTrigger`, we could have just skipped that for limited development time.

STL does not support sequence mode means, STL does not have public extension methods that support forcing voltage/current sequence (we do have an internal `ForceVoltageSequenceSynchronized` which needs more validation, and a public `ConfigureSequence` method).

But for unit testing, we won't use extension methods that STL exposes anyway. One thing you could try is to simply set the `Source.Mode` to `DCPowerSourceMode.Sequence` in your test. At the end of the day, the tests you're adding here is to make sure we don't call `SendSoftwareEdgeTrigger` for unsupported models which end up throwing errors. You don't have to make it a full valid sequence mode forcing.

With all that said, if we want to close this PR ASAP, I'm fine we skip unit testing for the changes in `SendSoftwareEdgeTrigger`, I'm fine we don't address the `Abort` issue I mentioned earlier too, given it's an existing issue.

Tanish-NI: Okay.
I was trying somethings to fix the error I mentioned above. The error I was getting because I was calling `sessionsBundle.ConfigureTriggerSoftwareEdge(triggerType)` as well along with `Initiate` but once I removed that, `SendSoftwareEdgeTrigger` is working for `StartTrigger` now.
I have improved unit testing by adding more unit tests and making them `[Theory]` based to test for each trigger type and for each method I touched in this PR.

Also, I addressed the `Abort` issue in this PR itself and added conditional logic in all 4 instances we are calling `Abort` in `TriggerAndEvents.cs` file.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
Tanish-NI: Updated.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -115,11 +117,16 @@ public static void DisableTriggers(this DCPowerSessionsBundle sessionsBundle)
...
-                sessionInfo.ConfigurePulseTriggerDisable(pinSiteInfo.ModelString, pinSiteInfo.IndividualChannelString);
-                sessionInfo.ConfigureSequenceAdvanceTriggerDisable(pinSiteInfo.ModelString, pinSiteInfo.IndividualChannelString);
-                sessionInfo.ConfigureSourceTriggerDisable(pinSiteInfo.ModelString, pinSiteInfo.IndividualChannelString);
-                sessionInfo.ConfigureStartTriggerDisable(pinSiteInfo.ModelString, pinSiteInfo.IndividualChannelString);
+                var triggerTypesUnsupported = GetUnsupportedTriggerTypes(pinSiteInfo.ModelString);
+                var triggerTypesToDisable = new List<TriggerType>() { TriggerType.PulseTrigger, TriggerType.SequenceAdvanceTrigger, TriggerType.SourceTrigger, TriggerType.StartTrigger };
+                if (triggerTypesToDisable.Except(triggerTypesUnsupported).Any())
+                {
+                    sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Control.Abort();
+                    sessionInfo.ConfigurePulseTriggerDisable(pinSiteInfo.ModelString, pinSiteInfo.IndividualChannelString);
```
lornazh, line 125: Now that I think it again, I feel we could get rid of all the `ConfigureXXXTriggerDisable` sub functions, and even the `DoForSupportedModels` and `IsTriggerTypeSupported` methods by reusing the `triggerTypesUnsupported` variable value and hence checking the unsupported list just once for one extension method.

```
sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
{
  var triggerTypesUnsupported = GetUnsupportedTriggerTypes(pinSiteInfo.ModelString);
  var triggerTypesToDisable = new List<TriggerType>() { TriggerType.PulseTrigger, TriggerType.SequenceAdvanceTrigger, TriggerType.SourceTrigger, TriggerType.StartTrigger };
  var supportedTriggerTypesToDisable = triggerTypesToDisable.Except(triggerTypesUnsupported);
  if (supportedTriggerTypesToDisable.Any())
  {
    sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Control.Abort();
    if (supportedTriggerTypesToDisable.Contains(TriggerType.PulseTrigger))
    {
      sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Triggers.PulseTrigger.Disable();
    }
    if (supportedTriggerTypesToDisable.Contains(TriggerType.SequenceAdvanceTrigger))
    {
      sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Triggers.SequenceAdvanceTrigger.Disable();
    }
    if (supportedTriggerTypesToDisable.Contains(TriggerType.SourceTrigger))
    {
      sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Triggers.SourceTrigger.Disable();
    }
    if (supportedTriggerTypesToDisable.Contains(TriggerType.StartTrigger))
    {
      sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Triggers.StartTrigger.Disable();
    }
  }
}
```

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/TriggersAndEvents.cs
```diff
@@ -146,26 +153,31 @@ public static void ConfigureTriggerDigitalEdge(this DCPowerSessionsBundle sessio
             // Hence, need the ability to check the operation against each channel when configuring triggers.
             sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
             {
-                sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Control.Abort();
-                switch (triggerType)
+                var triggerTypesUnsupported = GetUnsupportedTriggerTypes(pinSiteInfo.ModelString);
+                var triggerTypesToConfigure = new List<TriggerType>() { TriggerType.MeasureTrigger, TriggerType.PulseTrigger, TriggerType.SequenceAdvanceTrigger, TriggerType.SourceTrigger, TriggerType.StartTrigger };
+                if (triggerTypesToConfigure.Except(triggerTypesUnsupported).Any())
```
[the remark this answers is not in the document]
Tanish-NI: Did all the refactoring suggested. Removed all ConfigureXXX APIs, removed DoForSupportedModels  method as well as IsTriggerTypeSupported utility method as suggested in below comment. 
Please have a look.

lornazh (approved):

Mattjet27 (approved):

PR #308, opened 2025-07-28
Updated Simulation logic for reset and write digital data and updated assembly name

hcnagaraj-e:
Updated write digital data and reset for simulation class.
Also updated assembly name and uploaded new assembly and uploaded sequence file consuming new assembly

Simulation code used in write digital data and reset does not seem to be working. Need to update the code logic.
Addressing new comments from previous PR https://github.com/ni/semi-test-library-dotnet/pull/306

Verified building and running example sequence. It works fine as expected.


Examples/source/Sequence/CustomInstrument/Code Modules/Imports/MyCustomInstrumentDriverAPI/source/Simulation.cs
```diff
@@ -5,42 +5,48 @@ namespace MyCompany.MyCustomInstrumentDriverAPI
...
         internal static void WriteDigitalChannelData(string instrumentName, string channelName, double data)
         {
-            SimulatedInstruments[instrumentName].DigitalInputChannelData[channelName] = data;
+            _simulatedInstruments[instrumentName].DigitalInputChannelData[channelName] = data;
         }
 
         internal static void WriteDigitalData(string instrumentName, double data)
         {
-            SimulatedInstruments[instrumentName].DigitalInputChannelData.Select(chVal => data);
+            foreach (var channelName in _simulatedInstruments[instrumentName].DigitalInputChannelData.Keys.ToList())
```
lornazh, line 30: I think `ToList` is not necessary.

hcnagaraj-e: @lornazh I actually tried without using `ToList`, first, it throws exception as we are updating the same dictionary. Adding `ToList` will avoid that situation, and it works fine

nitamil (approved):

PR #309, opened 2025-07-29
Updated limit ranges for runs on actual Tester

hcnagaraj-e:

1. Updated limit ranges for no-offline mode runs. 
2. Added `DAQ_4468_C2_S13` to `MixedSignal.offlinecfg` ( _**Could not find required slot, used different available slot**_)
3. Updated `AcquireAnalogInputWaveformsTests` to use `DAQmxMultiChannelTests.pinmap` which has `DAQ_4468_C2_S13`.

Published values for voltage values for tests `Continuity` and `ForceCurrentAndMeassureVoltage` tests, are very close to voltage limits set for tests on actual tester. Tester has actual SMU instruments with open circuit, since current cannot reach requested values, applied voltage keeps increasing till it reaches voltage limits used.
Need to update voltage limit range for these tests as per voltage limits used in the test.

Also, `DAQ_6363_C1_S08` is not available in LungYuan, only DAQ device available in LungYuan is `DAQ_4468_C2_S13`.
Need to update tests to use different PinMap file which contains `DAQ_4468_C2_S13`.
But `DAQ_4468_C2_S13` is not available in `MixedSignal.offlinecfg`. So, need to update offline config to include new device.

Tested on a VM with offline mode using MixedSignal.offlinecfg.
Tested on LungYuan tester, verified values are as per the limits.
#### Known issue
Newly added device in offline config, does not follow naming convention. Not sure what is the best way to handle


lornazh:
@hcnagaraj-e Do these tests get run in offline pipeline or ATS pipeline? I ask because I don't see any DAQmx specific pinmap file under the [LungYuan pinmap folder](https://ni.visualstudio.com/DevCentral/_git/srd-sebu-mixed-signal-apt?path=/SemiconductorTestLibraryDotNET/TestAssets/Supporting%20Materials/Pin%20Maps/LungYuan).

If you want them to be run in both offline pipeline and ATS pipeline (or to be more precise, in both simulated environment and LungYuan environment), then you need to create a separate pinmap file with the same name in LungYuan pinmap folder. In the pinmap for LungYuan, the resource name must match hardware setup (in your case `DAQ_4468_C2_S13`). In the pinmap for offline run, since the slot C2_S13 is already occupied, you should use a resource name that matches the slot you add the instrument to in the offline config file.

lornazh (changes requested):
Please do it in a way that complies with the auto test infrastructure.

lornazh (approved):

Mattjet27 (approved):

Tanish-NI (approved):

lornazh:
> > @lornazh @Mattjet27 I have added traits for remaining tests and included all 3 testers we use. Also, updated limits to have 0.01 Volts of tolerance. Assert fails if the actual value is exact as that of limits mentioned in the range. I have tested running this branch in LungYuan, tests are passing.
> 
> I have run the pipeline on LungYuan, all tests are running and passing now. [Functionality ATS TesStandSteps pipeline build](https://dev.azure.com/ni/DevCentral/_build/results?buildId=15056352&view=logs&j=9316236d-33d1-5e32-58fc-bdb37708dc8b&t=3eec813a-7de8-55d6-5d1a-710adf867018)
> 
> image
> Since, it was minor change, I am not proceeding with merge. Please let me know if any changes needed.

Looks good to me, no further changes needed.

PR #339, opened 2025-09-23
Merge Measure Bug Fix Validation Tests

nitamil:

Added auto tests that validate the Merge Measure Bug Fix.  The Bug is that when pins are merged, the measurements that return the `PinSiteData` object must associate the result with the group name instead of the pin name. 

Additional auto tests for the Merge feature Measurement Bug. This test covers all the user workflow APIs. 

Tested locally on a VM with auto tests

### DIFF
--- a/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
+++ b/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
@@ -821,6 +821,7 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
                 }
             }
         }
+
         private DCPowerSessionsBundle MergeAndForceVoltage(string pinGroupName, out string primaryPin)
         {
             _tsmContext = CreateTSMContext("Merged_4163.pinmap");


SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -765,5 +789,12 @@ private void AssertAllChannelsHaveResult(PinSiteData<double> results)
                 }
             }
         }
+        private void AssertAllSitesHavePinGroupResult(PinSiteData<double> results, string pinGroup)
```
lornazh: 1. Suggest to rename the method to be `AssertResultAssociatedWithPinGroupName`.
2. Suggest to rename `pinGroup` to be `pinGroupName`.
3. I feel it might be worth it to also check there is no value associates with primary pin name.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -765,5 +821,28 @@ private void AssertAllChannelsHaveResult(PinSiteData<double> results)
...
+            sessionsBundle.ConfigureSourceDelay(0);
+            sessionsBundle.ForceVoltage(voltageLevel: 3.6, waitForSourceCompletion: true);
+            return sessionsBundle;
+        }
+
+        private void AssertResultAssociatedWithPinGroupName(PinSiteData<double> results, string pinGroup, string primaryPin)
+        {
+            foreach (var siteNumber in results.SiteNumbers)
+            {
+                Assert.NotEqual(0, results.GetValue(siteNumber, pinGroup));
```
lornazh: Testing value is confusing and misleading, it's better to use `public bool TryGetValue(int siteNumber, string pinName, out T value)`.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -765,5 +821,28 @@ private void AssertAllChannelsHaveResult(PinSiteData<double> results)
...
+                void Operation() => results.GetValue(siteNumber, primaryPin);
```
lornazh: This is too complicated, I think we could simply verify with `public bool TryGetValue(int siteNumber, string pinName, out T value)`.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -765,5 +821,26 @@ private void AssertAllChannelsHaveResult(PinSiteData<double> results)
                 }
             }
         }
+        private DCPowerSessionsBundle MergeAndForceVoltage(string pinGroupName, out string primaryPin)
```
lornazh: Please insert an empty line before this new method.

lornazh (approved):

PR #346, opened 2025-10-06
[POC1] Introducing deprecated overload methods to fix TestStand step warnings.

hcnagaraj-e:

1. Re-introduced `NIInstrumentType` enum in `SetupAndCleanupSteps` and marked obsolete.
2. Re-introduced 'CleanupInstrumentation' overload which accepts deprecated enum and marked the method as deprecated
3. Added new overload for 'CleanupInstrumentation' which just accepts only required parameters. (this will fix ambiguity issue)
4. Re-introduced 'ResetInstrumentation' with old prototype and marked it obsolete. Also added new overload without optional parameters to fix ambiguity issue.

5. Re-introduced `SetupNIDigitalPatternInstrumentation` overload which does not accept 4th optional parameter (`bool applySourceWaveformData = false`) and marked it obsolete.
6. Added new overload for `SetupNIDigitalPatternInstrumentation` which accepts only required parameter. (this will fix ambiguity issue)

POC created for supporting forward compatibility in Test Stand for STL

The enum `NIInstrumentType` was defined earlier inside `SetupAndCleanupSteps` call. Now it is moved out of `SetupAndCleanupSteps`class. When a test sequence created in 24.5 STL is upgraded to >25.0 and above, this creates a warning against `CleanupInstrumentation` step. The Sequence runs fine but user has to update the prototype by clicking on the warning.

Re-Introducing the deprecated overload method will allow old Sequence load without warning. Since, deprecated method does call back to other overload, sequence also runs fine.

Tested all upgrade downgrade scenarios, results can be found below. Upgrade scenarios are working fine. 
For downgrade scenarios are partially fixed, we get just the type mismatch pop-up and rest of things works fine.

image


SemiconductorTestLibrary.TestStandSteps/source/CleanupInstrumentation.cs
```diff
@@ -82,5 +83,89 @@ public static void CleanupInstrumentation(
...
+        }
+
+        /// <summary>
+        /// Closes any open instrument sessions associated with the pin map. To close instruments of specific types, use other overloads instead.
+        /// </summary>
+        /// <remarks>
+        /// This method makes a callback to new overload  method.
+        /// </remarks>
+        /// <param name="tsmContext">The <see cref="ISemiconductorModuleContext"/> object.</param>
+        public static void CleanupInstrumentation(ISemiconductorModuleContext tsmContext)
```
lornazh: Why do we need this overload method?

hcnagaraj-e: This is to remove ambiguity when just the `CleanupInstrumentation` method called without any optional parameters. All the autotests call these methods without optional parameters, it will break those tests without this overload.

It was suggested by Matt to fix the ambiguity problem.

SemiconductorTestLibrary.TestStandSteps/source/ResetInstrumentation.cs
```diff
@@ -77,5 +78,34 @@ public static void ResetInstrumentation(
...
+            tsmContext,
+            resetDevice,
+            (TestStandSteps.NIInstrumentType)instrumentType);
+        }
+
+        /// <summary>
+        /// Resets all the supported instrument sessions associated with the pin map. To reset specific type of instruments, use other overload instead.
+        /// </summary>
+        /// <param name="tsmContext">The <see cref="ISemiconductorModuleContext"/> object.</param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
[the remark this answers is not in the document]
hcnagaraj-e: I have removed [EditorBrowsable(EditorBrowsableState.Never)] for new overload

PR #354, opened 2025-11-20
[SMU Gang Pin Group] - Splitting Force Method to be able to configure and initiate channels independently

gmrsekar:
This PR contains changes related to splitting of `Force` method into ConfigureChannels and InitiateChannels therby providing a way to just configure channels and later initiating it.

In order to gang channels, leader and follower channels must be configured first, follower channels must then be initiated and would be waiting for trigger from the leader channel. Finally leader channel must be initiated which would send the trigger to follower channels and all the ganged channels are initiated at the same time. Hence configuring and initiating channels must be separated and should happen independently. 

`ConfigureChannels` and `InitiateChannels` methods would be used in `ForceCurrent`, `ForceVoltage` and all their overloads to support ganging which will be part of upcoming PRs.

For normal operations, `Force` method is directly called and hence there won't be any impact by this change.

As part of ganging implementation, `SitePinInfo` object which holds the ganging related properties through its `CascadingInfo` member needs to be accessed while configuring the channels with ganging related properties. Hence `Force` method's signature is changed to include `sitePinInfo` as optional argument. Optional parameter `channelString` is removed because channel string can be obtained from `sitePinInfo` itself.

So, when `Force` method is called with `sitePinInfo` as argument, forcing happens on the particular channel associated to that `sitePinInfo`. If not, all channels present in the session information are forced.

`ConfigureChannels` and `InitiateChannels` both take `channelOutput` as arguement.
`ConfigureSourceSettings` method's signature will also be modified in upcoming PR to include `sitePinInfo` and `channelOutput` and remove `channelString`.

All the references of `Force` method are updated to reflect the change in signature.

No new functionality has been added and `Force` is a private method. Hence new auto tests haven't been created.  The existing ones all are passing, ensuring that the changes in the PR aren't breaking any existing functionality.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -320,7 +320,7 @@ public static void ForceCurrent(this DCPowerSessionsBundle sessionsBundle, IDict
                     LevelRange = currentLevelRange,
                     LimitRange = voltageLimitRange
                 };
-                sessionInfo.Force(settings, sitePinInfo.IndividualChannelString, waitForSourceCompletion: waitForSourceCompletion);
+                sessionInfo.Force(settings, sitePinInfo, waitForSourceCompletion: waitForSourceCompletion);
```
lornazh: Not your fault, but named parameter does not seem necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -347,7 +347,7 @@ public static void ForceCurrent(this DCPowerSessionsBundle sessionsBundle, SiteD
...
```
lornazh: Not your fault, but named parameter does not seem necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -374,7 +374,7 @@ public static void ForceCurrent(this DCPowerSessionsBundle sessionsBundle, PinSi
...
```
lornazh: Not your fault, but named parameter does not seem necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -390,7 +390,7 @@ public static void ForceCurrent(this DCPowerSessionsBundle sessionsBundle, DCPow
             sessionsBundle.Do(sessionInfo =>
             {
                 settings.OutputFunction = DCPowerSourceOutputFunction.DCCurrent;
-                sessionInfo.Force(settings, sessionInfo.AllChannelsString, waitForSourceCompletion);
+                sessionInfo.Force(settings, waitForSourceCompletion: waitForSourceCompletion);
```
lornazh: Named parameter does not seem necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -515,7 +515,7 @@ internal static void ForceVoltageSequenceSynchronized(
                 };
                 // Applies limits and ranges.
                 var perChannelString = sitePinInfo.IndividualChannelString;
-                sessionInfo.Force(settings, perChannelString);
+                sessionInfo.Force(settings, sitePinInfo);
```
lornazh: Since the `Force` method on line 518 does not use `perChannelString` anymore, suggest moving line 517 to before line 520 where it's first used.

lornazh (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -935,12 +935,24 @@ public static void ConfigureSourceSettings(this DCPowerSessionInformation sessio
 
         #region private and internal methods
 
-        private static void Force(this DCPowerSessionInformation sessionInfo, DCPowerSourceSettings settings, string channelString = "", bool waitForSourceCompletion = false)
+        private static void Force(this DCPowerSessionInformation sessionInfo, DCPowerSourceSettings settings, SitePinInfo sitePinInfo = null, bool waitForSourceCompletion = false)
+        {
+            var channelString = sitePinInfo?.IndividualChannelString ?? sessionInfo.AllChannelsString;
+            var channelOutput = sessionInfo.Session.Outputs[channelString];
+            sessionInfo.ConfigureChannels(settings, channelOutput, channelString, sitePinInfo);
```
gmrsekar: Line 942 and 949 to be changed once the signature of `ConfigureSourceSettings` method is finalized and the [PR](https://github.com/ni/semi-test-library-dotnet/pull/354) is merged into main.

PR #358, opened 2025-11-24
Implementation of Force Current Sequencing and its helper methods

Ashutosh-NI:
This pull request adds comprehensive support for hardware-timed current source sequencing to the DC Power abstraction layer, including per-site and per-pin configurations, as well as per-step source delays. It also introduces new unit tests to validate these features. The main changes are grouped below:

### New API Methods for Current Sequencing

* Added three overloads of `ForceCurrentSequence` to `DCPowerSessionsBundle` for applying current sequences: one for all channels, one for per-site (`SiteData<double[]>`), and one for per-pin-per-site (`PinSiteData<double[]>`). 

### Sequence Configuration Enhancements

* Added overloads of `ConfigureSequence` to `DCPowerSessionsBundle` for per-site and per-pin-per-site support, enabling flexible configuration of voltage/current sequences with loop counts and step timing.
* Added overloads of `ConfigureSequenceWithSourceDelays` to `DCPowerSessionsBundle` for all channels, per-site, and per-pin-per-site support, allowing configuration of per-step source delays in sequences.
* Added a new method to `DCPowerOutput` for configuring sequences with per-step source delays, integrating time spans for each step.

### Unit Test Coverage

* Added multiple unit tests to `SourceTests.cs` to verify correct behavior of the new sequence APIs, including tests for per-site and per-pin-per-site data, compliance limits, and correct value settings for various hardware configurations.
* Added a new assertion helper to validate current sequence settings, including checks for current level range, voltage limits, and sequence loop count.

For more details : [HLD](https://dev.azure.com/ni/DevCentral/_git/AppCentral.wiki?path=/Organizations/Systems-R%26D/Groups/Semiconductor-(SEBU)/Semi-APT-App-SW/Mixed-Signal-APT/Feature-Specs/26.0/%5BHLD%5D-DC-Power-Hardware-Level-Sequence-Support-in-STL.md&version=GBwikiMaster&_a=preview&anchor=simple-sequence-methods)

### DIFF
--- a/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
+++ b/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
@@ -14,6 +14,8 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCP
     /// </summary>
     public static class Source
     {
+        private const double DefaultSequenceTimeOut = 5;
+
         #region methods on DCPowerSessionsBundle
 
         /// <summary>
@@ -493,7 +495,7 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCP
             double? voltageLimitRange = null,
             int sequenceLoopCount = 1,
             bool waitForSequenceCompletion = false,
-            double sequenceTimeoutInSeconds = 5.0)
+            double sequenceTimeoutInSeconds = DefaultSequenceTimeOut)
         {
             sessionsBundle.Do(sessionInfo =>
             {
@@ -519,7 +521,7 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCP
             double? voltageLimitRange = null,
             int sequenceLoopCount = 1,
             bool waitForSequenceCompletion = false,
-            double sequenceTimeoutInSeconds = 5.0)
+            double sequenceTimeoutInSeconds = DefaultSequenceTimeOut)
         {
             sessionsBundle.Do((sessionInfo, sitePinInfo) =>
             {
@@ -549,7 +551,7 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCP
             double? voltageLimitRange = null,
             int sequenceLoopCount = 1,
             bool waitForSequenceCompletion = false,
-            double sequenceTimeoutInSeconds = 5.0)
+            double sequenceTimeoutInSeconds = DefaultSequenceTimeOut)
         {
             sessionsBundle.Do((sessionInfo, sitePinInfo) =>
             {
@@ -766,22 +768,22 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCP
         /// <inheritdoc cref="ConfigureSequence(DCPowerSessionsBundle, double[], int, double?)"/>
         public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle, SiteData<double[]> sequence, int sequenceLoopCount = 1, double? sequenceStepDeltaTimeInSeconds = null)
         {
-            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
             {
-                var channelOutput = sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString];
+                var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
                 channelOutput.Control.Abort();
-                channelOutput.ConfigureSequence(sequence.GetValue(pinSiteInfo.SiteNumber), sequenceLoopCount, sequenceStepDeltaTimeInSeconds);
+                channelOutput.ConfigureSequence(sequence.GetValue(sitePinInfo.SiteNumber), sequenceLoopCount, sequenceStepDeltaTimeInSeconds);
             });
         }
 
         /// <inheritdoc cref="ConfigureSequence(DCPowerSessionsBundle, double[], int, double?)"/>
         public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle, PinSiteData<double[]> sequence, int sequenceLoopCount = 1, double? sequenceStepDeltaTimeInSeconds = null)
         {
-            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
             {
-                var channelOutput = sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString];
+                var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
                 channelOutput.Control.Abort();
-                channelOutput.ConfigureSequence(sequence.GetValue(pinSiteInfo), sequenceLoopCount, sequenceStepDeltaTimeInSeconds);
+                channelOutput.ConfigureSequence(sequence.GetValue(sitePinInfo), sequenceLoopCount, sequenceStepDeltaTimeInSeconds);
             });
         }
 
--- a/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
+++ b/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
@@ -659,12 +659,12 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
 
             sessionsBundle.ConfigureMeasureWhen(DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete);
             var sequence = new SiteData<double[]>(new double[][]
-                        {
-                            new[] { -0.005, 0.010 },  // Site 0
-                            new[] { -0.006, 0.012 },  // Site 1
-                            new[] { 0.007, 0.014 },  // Site 2
-                            new[] { 0.008, 0.016 }
-                        });
+            {
+                new[] { -0.005, 0.010 },
+                new[] { -0.006, 0.012 },
+                new[] { 0.007, 0.014 },
+                new[] { 0.008, 0.016 }
+            });
             sessionsBundle.ForceCurrentSequence(currentSequence: sequence, voltageLimit: 0.5, currentLevelRange: 0.1, voltageLimitRange: 1, sequenceLoopCount: 2);
 
             sessionsBundle.Abort();
@@ -1144,7 +1144,7 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
         [Theory]
         [InlineData(false)]
         [InlineData(true)]
-        public void DifferentSMUDevices_SetSequenceWithSourceDelaySucceeds(bool pinMapWithChannelGroup)
+        public void DifferentSMUDevices_ConfigureSequenceWithSourceDelaySucceeds(bool pinMapWithChannelGroup)
         {
             var sessionManager = Initialize(pinMapWithChannelGroup);
             var sessionsBundle = sessionManager.DCPower("VDD");


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -474,6 +474,160 @@ public static void ForceCurrentAsymmetricLimit(this DCPowerSessionsBundle sessio
...
+                    sequenceLoopCount,
+                    waitForSequenceCompletion,
+                    sequenceTimeoutInSeconds);
+            });
+        }
+
+        /// <summary>
+        /// Core implementation for forcing a current sequence on a specific channel.
+        /// </summary>
+        private static void ForceCurrentSequenceCore(
```
gmrsekar: Can we have the name as `ForceSequence` if it will be common to both voltage and current sequencing? Because we have `ForceCurrent` and `ForceVoltage` as extension methods on bundle and `Force` as extension method on sessionInformation and `ForceSequence` will be synonymous to that.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -606,6 +606,89 @@ public void DifferentSMUDevices_ForceCurrentWithAsymmetricLimitAndRangesSucceeds
...
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void DifferentSMUDevices_ForceSequenceCurrentAndSucceeds(bool pinMapWithChannelGroup)
+        {
+            var sessionManager = Initialize(pinMapWithChannelGroup);
+            var sessionsBundle = sessionManager.DCPower("VDD");
+            var sequence = new[] { 0.000, 0.005, 0.010 };
+
+            sessionsBundle.ForceCurrentSequence(sequence);
```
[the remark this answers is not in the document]
Ashutosh-NI: The idea of this test to just check if the `ForceCurrentSequence` is running without any exception , that is why this test does not have any assert.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -474,6 +474,150 @@ public static void ForceCurrentAsymmetricLimit(this DCPowerSessionsBundle sessio
             });
         }
 
+        /// <summary>
+        /// Forces a hardware-timed sequence of current values on the targeted pin(s).
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
+        /// <param name="currentSequence">Array of current levels (in amperes) to source step-by-step. First element is applied before initiating the sequence.</param>
```
nitamil: This line is not required to be mentioned here.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -474,6 +474,150 @@ public static void ForceCurrentAsymmetricLimit(this DCPowerSessionsBundle sessio
...
+        /// <param name="voltageLimit">Optional symmetric compliance voltage limit for the sequence (in volts).</param>
+        /// <param name="currentLevelRange">Optional current level range (in amperes). Defaults to |currentSequence[i]| if not specified.</param>
```
nitamil: Please use the word `absolute value of` instead of symbol.

nitamil: Also, are we setting this property for every set point.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -1018,6 +1126,93 @@ public void DifferentSMUDevices_SetSequenceWithStepDeltaTime_SequenceStepDeltaTi
...
+                new[] { 1.5, 2.5, 3.5, 4.5, 5.5 },  // Site 1
+                new[] { 2.0, 3.0, 4.0, 5.0, 6.0 },  // Site 2
+                new[] { 2.5, 3.5, 4.5, 5.5, 6.5 } // Site 3
+            });
+            var expectedSourceDelayInSeconds = new SiteData<double[]>(new double[][]
+            {
+                new[] { 5.0, 5.0, 5.0, 5.0, 5.0 },  // Site 0
+                new[] { 5.0, 5.0, 5.0, 5.0, 5.0 },  // Site 1
+                new[] { 5.0, 5.0, 5.0, 5.0, 5.0 },  // Site 2
+                new[] { 5.0, 5.0, 5.0, 5.0, 5.0 }, // Site 3
```
nitamil: Uneven spacing between the comments on previous lines and this line. Applies to previous variable as well.

nitamil (approved):
Approved with comments.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -474,6 +474,131 @@ public static void ForceCurrentAsymmetricLimit(this DCPowerSessionsBundle sessio
...
+                    voltageLimitRange,
+                    sequenceLoopCount,
+                    waitForSequenceCompletion,
+                    sequenceTimeoutInSeconds,
+                    DCPowerSourceOutputFunction.DCCurrent);
+            });
+        }
+
+        /// <summary>
+        /// Core implementation for forcing a current/voltage sequence on a specific channel.
```
lornazh: The description is not precise because this method handles multiple channels too.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -990,6 +1211,10 @@ internal static bool IsFirstChannelOfSession(this SitePinInfo sitePinInfo, DCPow
 
         private static void ConfigureVoltageSettings(DCPowerOutput dcOutput, DCPowerSourceSettings settings)
         {
+            if (settings.OutputFunction.HasValue)
```
lornazh: Maybe we could directly set output function to Voltage since the `OutputFunction` property must have a value and the value must be Voltage for us to come here?

By the way, will the `Equals` check in the caller of this method return false or throw an error when `OutputFunction` is null?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -614,6 +614,81 @@ public void DifferentSMUDevices_ForceCurrentWithAsymmetricLimitAndRangesSucceeds
             sessionsBundle.ForceCurrentAsymmetricLimit(currentLevel: 0.1, voltageLimitHigh: 3, voltageLimitLow: -1, currentLevelRange: 0.5, voltageLimitRange: 5);
         }
 
+        [Theory]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
```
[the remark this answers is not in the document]
lornazh: Ah okay, I mis-remember that "do not filter for offline ATS" part then (it IS a little confusing...)

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
Ashutosh-NI: >Maybe we could directly set output function to Voltage since the OutputFunction property must have a value and the value must be Voltage for us to come here?

yes this makes more sense! will make the similar kind of change for the `ConfigureCurrentSettings` as well.


>By the way, will the Equals check in the caller of this method return false or throw an error when OutputFunction is null?

The `Equals` check returns false if value does not matches.
Image

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -474,6 +474,131 @@ public static void ForceCurrentAsymmetricLimit(this DCPowerSessionsBundle sessio
...
+                    DCPowerSourceOutputFunction.DCCurrent);
+            });
+        }
+
+        /// <summary>
+        /// Core implementation for forcing a current/voltage sequence on a specific channel.
+        /// </summary>
+        private static void ForceSequenceCore(
+            DCPowerOutput channelOutput,
+            double[] currentSequence,
```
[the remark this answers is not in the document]
Ashutosh-NI: updated!

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -638,6 +763,38 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
...
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
+        /// <param name="sequence">The voltage or current sequence to set.</param>
+        /// <param name="sequenceLoopCount">The number of loops a sequence runs after initiation.</param>
+        /// <param name="sequenceStepDeltaTimeInSeconds">The delta time between the start of two consecutive steps in a sequence.</param>
+        public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle, SiteData<double[]> sequence, int sequenceLoopCount = 1, double? sequenceStepDeltaTimeInSeconds = null)
+        {
+            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
+            {
+                sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Control.Abort();
```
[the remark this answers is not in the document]
Ashutosh-NI: updated!

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -898,10 +1108,6 @@ public static void ConfigureSourceSettings(this DCPowerSessionInformation sessio
             {
```
lornazh: I believe you need to move the setting of `ComplianceLimitSymmetry` to the new `ConfigureLevelsAndLimits` method, right?

Set PR to waiting mainly for this comment.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -990,6 +1201,7 @@ internal static bool IsFirstChannelOfSession(this SitePinInfo sitePinInfo, DCPow
 
         private static void ConfigureVoltageSettings(DCPowerOutput dcOutput, DCPowerSourceSettings settings)
         {
+            dcOutput.Source.Output.Function = settings.OutputFunction.Value;
```
lornazh: I feel `dcOutput.Source.Output.Function = DCPowerSourceOutputFunction.DCVoltage` makes most sense.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -1222,5 +1379,38 @@ private static int[] GetActiveSites(DCPowerSessionsBundle sessionsBundle)
...
+                sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString].Control.Initiate();
+                return sessionInfo.Session.Measurement.Fetch(sitePinInfo.IndividualChannelString, PrecisionTimeSpan.FromSeconds(timeoutSeconds), itemsToFetch);
+            });
+
+            for (int i = 0; i < siteCount; i++)
+            {
+                var expectedSequence = getExpectedSequence(i);
+                if (checkForCurrentMeasurement)
+                {
+                    Assert.All(
```
lornazh: Can we simply use `Assert.Equal(expectedSequence, results[i][0].CurrentMeasurements, precision)`? I think `Equal` does a sequence comparison for arrays, checking each element for equality.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -795,6 +942,54 @@ public static void ConfigureSourceDelay(this DCPowerSessionsBundle sessionsBundl
...
+        /// <inheritdoc cref="ConfigureSequenceWithSourceDelays(DCPowerSessionsBundle, double[], double[], int)"/>
+        public static void ConfigureSequenceWithSourceDelays(
+            this DCPowerSessionsBundle sessionsBundle,
+            SiteData<double[]> sequence,
+            SiteData<double[]> sourceDelaysInSeconds,
+            int sequenceLoopCount = 1)
+        {
+            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
+            {
+                sessionInfo.Session.Outputs[pinSiteInfo.IndividualChannelString].Control.Abort();
```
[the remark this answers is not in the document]
Ashutosh-NI: updated the code!

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
Ashutosh-NI: Yes,  i tried using `Assert.Equal` initially, but `Assert.Equal` does not have precision parameter for double array comparison. It only have `comparer`. So i had to use it this way where I can convert them to single value.

Image

lornazh: Oh alright. I think there is still improvements we could make.

1. Create a variable `actualSequence = checkForCurrentMeasurement ? results[i][0].CurrentMeasurements : results[i][0].VoltageMeasurements` so that we don't have to repeat the complicated `Assert.All` logic twice.
2. We could create our own `AssertEqualWithPricisionDigits` utility to wrap the complicated logic for later usages.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -638,6 +763,28 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
             });
         }
 
+        /// <inheritdoc cref="ConfigureSequence(DCPowerSessionsBundle, double[], int, double?)"/>
+        public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle, SiteData<double[]> sequence, int sequenceLoopCount = 1, double? sequenceStepDeltaTimeInSeconds = null)
+        {
+            sessionsBundle.Do((sessionInfo, pinSiteInfo) =>
```
lornazh: `sitePinInfo`?

PR #362, opened 2025-11-26
Implementation for CreateRampSequence and its overloads

Ashutosh-NI:
This pull request adds new ramp sequence generation utilities to the `HelperMethods` class and introduces comprehensive unit tests to validate their behavior. The main changes are the addition of methods for creating ramp sequences and wrapping them in domain-specific data structures, as well as making `HelperMethods` public for broader accessibility.

* Added `CreateRampSequence` method to `HelperMethods` to generate an array of double values forming a ramp from a start value to a stop value with a specified number of points.
* Added overloads to wrap the generated ramp sequence in `SiteData<double[]>` and `PinSiteData<double[]>` objects for use with site numbers and pin names, respectively.

* Changed `HelperMethods` from `internal` to `public` and added XML documentation comments to improve discoverability and maintainability.

* Added `HelperMethodsTests` covering ramp sequence creation, including edge cases, descending ramps, and correct association with sites and pins.


nitamil (approved):

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +21,55 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
         }
+
+        /// <summary>
+        /// Creates a ramp sequence of double values from outputStart to outputStop with the specified number of points.
+        /// </summary>
+        /// <param name="outputStart">The starting value of the ramp.</param>
+        /// <param name="outputStop">The ending value of the ramp.</param>
+        /// <param name="numberOfPoints">The number of points in the ramp sequence.</param>
+        /// <returns>An array of double values representing the ramp sequence.</returns>
+        public static double[] CreateRampSequence(double outputStart, double outputStop, double numberOfPoints)
```
hcnagaraj-e: @Ashutosh-NI when I was working for PRnD products, we were advice to use 'AAL libraries' (Arithmetic and Logic math libraries maintained by dedicated team) for all complex mathematical operations like trajectory generations. I am not sure if AAL has C Sharp libraries. Those libraries are well tested for all possible inputs and corner cases.

@lornazh What is your recommendation, should we check with AAL team?

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +21,55 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
+        {
+            double stepSize = 0.0;
+
+            if (numberOfPoints > 1)
+            {
+                stepSize = (outputStop - outputStart) / (numberOfPoints - 1);
```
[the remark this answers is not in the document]
Ashutosh-NI: line 38 will only get executed when numberOfPoints > 1 as it is inside conditional block

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +21,55 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
+            }
+            double[] rampSequence = new double[(int)numberOfPoints];
+            for (int i = 0; i < numberOfPoints; i++)
+            {
+                rampSequence[i] = outputStart + (i * (double)stepSize);
```
lornazh: Will there be tiny rounding drift which makes the last element in `rampSequence` not equal to `outputStop`?

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +21,55 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
+
+        /// <summary>
+        /// Creates a ramp sequence and wraps it in a SiteData object for the specified site numbers.
+        /// </summary>
+        /// <param name="outputStart">The starting value of the ramp.</param>
+        /// <param name="outputStop">The ending value of the ramp.</param>
+        /// <param name="numberOfPoints">The number of points in the ramp sequence.</param>
+        /// <param name="siteNumbers">The site numbers to associate with the ramp sequence.</param>
+        /// <returns>A SiteData object containing the ramp sequence for the specified sites.</returns>
+        public static SiteData<double[]> CreateRampSequence(double outputStart, double outputStop, int numberOfPoints, int[] siteNumbers)
```
lornazh: Would it be better to let `siteNumbers` go first?

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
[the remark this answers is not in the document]
Ashutosh-NI: Yes, we should.
What do you suggest, should we throw exception if above cases comes? or should we use the absolute value(`|outputStop - outputStart|`) to continue the ramp generation?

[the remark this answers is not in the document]
lornazh: Let's throw exception. We'd better not hiding user mistakes and workaround it with some non-intuitive way.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.resx
```diff
@@ -138,4 +138,13 @@
   <data name="Digital_TimeSetPeriodNotDistinct" xml:space="preserve">
     <value>The value of the time set period ({0}) is not the same for all underlying instrument sessions.</value>
   </data>
+  <data name="InvalidNumberOfPoints" xml:space="preserve">
```
lornazh: The name is a little generic, can we add a prefix like other names, for example `CreateRamp_InvalidNumberOfPoints`?

SemiconductorTestLibrary.Extensions/source/ResourceStrings.resx
```diff
@@ -138,4 +138,13 @@
   <data name="Digital_TimeSetPeriodNotDistinct" xml:space="preserve">
     <value>The value of the time set period ({0}) is not the same for all underlying instrument sessions.</value>
   </data>
+  <data name="InvalidNumberOfPoints" xml:space="preserve">
+    <value>Number of points must be greater than zero. </value>
+  </data>
+  <data name="InvalidOutPutStart" xml:space="preserve">
```
lornazh: "OutPut" should be one word.

SemiconductorTestLibrary.Extensions/tests/Common/HelperMethodsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+using Xunit;
+
+namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.Common
+{
+    public class HelperMethodsTests
+    {
+        [Theory]
+        [InlineData(0.0, 1.0, 5)]
+        [InlineData(-1.0, 1.0, 3)]
+        [InlineData(2.5, 2.5, 1)]
```
lornazh: Could be removed since it's covered in `CreateRampSequence_SinglePointRampProducesConstant`.

SemiconductorTestLibrary.Extensions/tests/Common/HelperMethodsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        public void CreateRampSequence_RampValuesAreCorrect(double start, double stop, int points)
```
lornazh: I actually like the ones below (e.g. `CreateRampSequence_DescendingRampValuesCorrect`) more than this one because they are very specific about the cases to verify.

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +22,75 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
+        /// <param name="numberOfPoints">The number of points in the ramp sequence.</param>
+        /// <returns>An array of double values representing the ramp sequence.</returns>
+        /// <exception cref="ArgumentException">Thrown when numberOfPoints is less than or equal to zero.</exception>
+        /// <exception cref="ArgumentException">Thrown when outputStart is NaN or Infinity.</exception>
+        /// <exception cref="ArgumentException">Thrown when outputStop is NaN or Infinity.</exception>
+        public static double[] CreateRampSequence(double outputStart, double outputStop, int numberOfPoints)
+        {
+            if (numberOfPoints <= 0)
+            {
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.InvalidNumberOfPoints));
```
lornazh: Not sure if we want to throw an STL exception or a generic argument exception.

SemiconductorTestLibrary.Extensions/tests/Common/HelperMethodsTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            var seq = HelperMethods.CreateRampSequence(5.0, 1.0, 5);
+
+            Assert.Equal(new[] { 5.0, 4.0, 3.0, 2.0, 1.0 }, seq);
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(-1)]
+        [InlineData(-10)]
+        public void CreateRampSequence_ZeroOrNegativeNumberOfPoints_ThrowsArgumentException(int numberOfPoints)
```
lornazh: There are only 2 parts in method logic, so maybe `CreateRampSequenceWithZeroOrNegativeNumberOfPoints_ThrowsArgumentException` is better?

SemiconductorTestLibrary.Extensions/source/Common/HelperMethods.cs
```diff
@@ -17,5 +22,75 @@ internal static T GetDistinctValue<T>(IEnumerable<T> values, string errorMessage
...
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.InvalidOutPutStart));
+            }
+            if (double.IsNaN(outputStop) || double.IsInfinity(outputStop))
+            {
+                throw new ArgumentException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.InvalidOutputStop));
+            }
+
+            double[] rampSequence = new double[numberOfPoints];
+
+            if (numberOfPoints == 1)
```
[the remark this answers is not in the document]
Ashutosh-NI: The idea for not throwing in above case was that the `numberOfPoints =1` depends on the `outputStart` and does not depend on the value for `outputStop`(eventough we might check for `outputStart == outputStop` but we will always return `outputStart` regardless `outputStop`) so throwing exception might add a restriction which is not that much required.

Please let me know your thoughts.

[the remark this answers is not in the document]
lornazh: > the numberOfPoints =1 depends on the outputStart and does not depend on the value for outputStop

So who decided that? Is that a common behavior? Will customers get that?

Even, I'm curious why would a customer want a ramp sequence with only one point? I feel it's more likely a typo, which we'd better report.

We could report `outputStart` and `outputStop` not equal, we may also require the `numberOfPoints` to be greater than 1 (instead of 0).

lornazh (approved):

PR #367, opened 2025-12-07
Add ForceVoltageSequence methods to DCPower Source

Ashutosh-NI:
This pull request introduces support for forcing hardware-timed voltage sequences on DC power supplies, along with comprehensive unit tests to verify the new functionality. The changes include new overloads of the `ForceVoltageSequence` method to support different data granularities (per array, per site, per pin-site), a core implementation method, and helper assertions for testing.

* Added three overloads of the `ForceVoltageSequence` method in `Source.cs` to allow forcing a sequence of voltage levels on all channels, per site, or per pin-site, with optional parameters for current limit, voltage range, loop count, and completion waiting.
* Implemented the private `ForceVoltageSequenceCore` method to centralize the logic for configuring and executing voltage sequences on the hardware.

* Added multiple unit tests in `SourceTests.cs` to verify correct application of voltage sequences for arrays, per-site, and per-pin-site data, as well as default parameter handling.
* Introduced the `AssertVoltageSequenceSettings` helper method to streamline assertions on voltage sequence settings in tests


hcnagaraj-e (approved):

nitamil (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -241,6 +241,100 @@ public static void ForceVoltage(this DCPowerSessionsBundle sessionsBundle, PinSi
             });
         }
 
+        /// <inheritdoc cref="ForceVoltageSequence(DCPowerSessionsBundle, double[], double?, double?, double?, int, bool, double)"/>
+        public static void ForceVoltageSequence(
```
lornazh: Can we reverse the order of these 3 methods so that it matches the order of overloads (single data -> SiteData -> PinSiteData) for other methods in this file?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -1214,6 +1316,18 @@ private void AssertCurrentSettings(DCPowerOutput channelOutput, double expectedC
             Assert.Equal(expectedVoltageLimitLow, channelOutput.Source.Current.VoltageLimitLow);
         }
 
+        private void AssertVoltageSequenceSettings(DCPowerOutput channelOutput, double? expectedCurrentLimit = null, double? expectedCurrentLimitRange = null)
```
lornazh: You are actually not asserting sequence settings, right? So maybe `AssertVoltageSettings` is a better name?

There are already 2 different `AssertVoltageSettings` overloads, we might want to think about how they should be used in different cases.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -324,6 +324,108 @@ public void DifferentSMUDevices_ForceVoltageWithAsymmetricLimitAndRangesSucceeds
             sessionsBundle.ForceVoltageAsymmetricLimit(voltageLevel: 3.6, currentLimitHigh: 0.2, currentLimitLow: -0.1, voltageLevelRange: 5, currentLimitRange: 0.5);
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public void DifferentSMUDevices_ForceVoltageSequenceArray_CorrectValueAreSet(bool pinMapWithChannelGroup)
```
lornazh: I don't thing the word "Array" is necessary because "Sequence" already indicates it's not a single value.

PR #379, opened 2026-01-05
Implement ConfigureAdvanceSequence and its overloads

Ashutosh-NI:
This pull request adds support for configuring advanced DC power sequences with per-step property settings. It introduces new overloads for the `ConfigureAdvancedSequence` method.

**Advanced Sequence Configuration:**

* Introduced multiple overloads of `ConfigureAdvancedSequence` in `Source.cs` to support:
  - Scalar list of step properties.
  - Per-site step properties using `SiteData`.
  - Per-pin-per-site step properties using `PinSiteData`.
  - All overloads funnel into a core implementation for consistency and maintainability. 

**Resource String Updates:**
* Added a new resource string, `DCPowerDeviceNotSupported`, for error messages when a property is not supported by a given instrument model.

**Unit Testing Improvements:**

* Added comprehensive unit tests in `SourceTests.cs` to verify:
  - Correct configuration and activation of advanced sequences for scalar, per-site, and per-pin-per-site step properties.
  - Validation of expected measurement results and sequence activation state.

### DIFF
--- a/Examples/source/CodeSnippets/InstrumentAbstraction/MeasureVoltageMixed.cs
+++ b/Examples/source/CodeSnippets/InstrumentAbstraction/MeasureVoltageMixed.cs
@@ -1,9 +1,9 @@
-﻿using NationalInstruments.SemiconductorTestLibrary.Common;
-using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction;
+﻿using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction;
 using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DCPower;
 using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.Digital;
 using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.DMM;
 using NationalInstruments.TestStand.SemiconductorModule.CodeModuleAPI;
+using static NationalInstruments.SemiconductorTestLibrary.Common.Utilities;
 
 namespace NationalInstruments.Examples.SemiconductorTestLibrary.CodeSnippets.InstrumentAbstraction
 {
@@ -30,7 +30,7 @@ namespace NationalInstruments.Examples.SemiconductorTestLibrary.CodeSnippets.Ins
             var dmmPins = sessionManager.DMM(filteredPinNamesPpmu);
 
             // Just an example of the InvokeInParallel method, assumes that the instrumentation is already configured.
-            Utilities.InvokeInParallel(
+            InvokeInParallel(
                 () => ppmuPins.MeasureAndPublishCurrent(publishedDataId),
                 () => smuPins.MeasureAndPublishCurrent(publishedDataId),
                 () => dmmPins.ReadAndPublish(maximumTimeInMilliseconds: 2000, publishedDataId));
@@ -48,7 +48,7 @@ namespace NationalInstruments.Examples.SemiconductorTestLibrary.CodeSnippets.Ins
             // Just an example of the InvokeInParallel method, assumes that the instrumentation is already configured.
             // Also, note that the null-conditional operator (?.) allows for the operation to be skipped when the SessionBundle object is null,
             // which is the case when no pins of that instrument type are passed into the function when using the filterPins parameter.
-            Utilities.InvokeInParallel(
+            InvokeInParallel(
                 () => ppmuPins?.MeasureAndPublishCurrent(publishedDataId),
                 () => smuPins?.MeasureAndPublishCurrent(publishedDataId),
                 () => dmmPins?.ReadAndPublish(maximumTimeInMilliseconds: 2000, publishedDataId));


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1081,6 +1087,82 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
...
+        }
+
+        private static void ConfigureAdvanceSequenceCore(
+            this DCPowerSessionsBundle sessionsBundle,
+            string sequenceName,
+            AdvancedSequenceProvider perStepProperties,
+            bool setAsActiveSequence,
+            bool commitFirstElementAsInitialState)
+        {
+            sessionsBundle.Do((DCPowerSessionInformation sessionInfo, SitePinInfo pinSiteInfo) =>
```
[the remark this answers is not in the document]
Ashutosh-NI: implemented

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -89,5 +93,1763 @@ public static IEnumerable<TriggerType> GetUnsupportedTriggerTypes(string modelSt
                 }
             }
         }
+
+        public static DCPowerAdvancedSequenceProperty[] ExtractAdvancedSequencePropertiesArray(
```
hcnagaraj-e: @Ashutosh-NI Is it possible to make use of property → enum mapping table? 

We can also use Codegen  to create this code at build time. For manual building, we need to include Codegen code in our dev machine setup script.  We used to do this for C++ code, not sure if it makes sense here.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1081,6 +1081,89 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
...
+            string sequenceName,
+            DCPowerOutput channelOutput,
+            string modelString,
+            IList<DCPowerAdvancedSequenceStepProperties> perStepProperties,
+            bool setAsActiveSequence,
+            bool commitFirstElementAsInitialState)
+        {
+            channelOutput.Source.Mode = DCPowerSourceMode.Sequence;
+            var advancedSequenceProperties = Utilities.ExtractAdvancedSequencePropertiesArray(perStepProperties);
+            channelOutput.Source.AdvancedSequencing.CreateAdvancedSequence(sequenceName, advancedSequenceProperties, true);
```
lornazh: Please use named parameter for hard coded bool values to improve readability.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -89,5 +93,1321 @@ public static IEnumerable<TriggerType> GetUnsupportedTriggerTypes(string modelSt
                 }
             }
         }
+
+        public static DCPowerAdvancedSequenceProperty[] ExtractAdvancedSequencePropertiesArray(IEnumerable<DCPowerAdvancedSequenceStepProperties> perStepProperties)
```
lornazh: Would `GetAdvancedSequencePropertiesToConfigure` be a better name?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -378,6 +378,173 @@ public void DifferentSMUDevices_ForceVoltageWithPerPinPerSiteSettingsObject_Corr
...
+            var sessionManager = Initialize(pinMapWithChannelGroup: false);
+            var sessionsBundle = sessionManager.DCPower("VDD");
+
+            sessionsBundle.ConfigureMeasureWhen(DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete);
+            var perSiteSteps = new SiteData<IList<DCPowerAdvancedSequenceStepProperties>>(new[]
+            {
+                new List<DCPowerAdvancedSequenceStepProperties>
+                {
+                    new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 1.0, OutputFunction = DCPowerSourceOutputFunction.DCVoltage },
+                    new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 1.2, OutputFunction = DCPowerSourceOutputFunction.DCVoltage }
```
lornazh: Do we want to test some other step properties combinations? Ideally we could test 3 different sets of step properties for these 3 unit tests.

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1081,6 +1081,89 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
...
+            }
+            for (int i = startIndex; i < perStepProperties.Count; i++)
+            {
+                channelOutput.Source.AdvancedSequencing.CreateAdvancedSequenceStep(true);
+                Utilities.ApplyStepProperties(channelOutput, perStepProperties[i], modelString);
+            }
+            if (!setAsActiveSequence)
+            {
+                channelOutput.Source.AdvancedSequencing.ActiveAdvancedSequence = string.Empty;
+            }
```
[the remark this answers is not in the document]
Ashutosh-NI: If we set `setAsActiveSequence` to false during the `CreateAdvancedSequence` call, the advanced sequence will not become active, and we won't be able to add any steps to it (steps can only be created when the sequence is active).
Therefore, for `ConfigureAdvanceSequence` to successfully create an advanced sequence and add steps to it, we must first mark the sequence as active. After the steps are added, we can then mark it as inactive if needed.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -89,5 +93,1321 @@ public static IEnumerable<TriggerType> GetUnsupportedTriggerTypes(string modelSt
...
+        {
+            var result = new HashSet<DCPowerAdvancedSequenceProperty>();
+
+            foreach (var stepProperties in perStepProperties)
+            {
+                foreach (var prop in typeof(DCPowerAdvancedSequenceStepProperties).GetProperties())
```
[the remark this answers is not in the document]
Ashutosh-NI: I think I can cache the `DCPowerAdvancedSequenceStepProperties.GetProperties()` also to reduce the time complexity furthermore.
With that said, do you want to benchmark it before completing the PR or should i take it as the part of testing user story?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1107,6 +1108,96 @@ public static void ConfigureSequence(this DCPowerSessionsBundle sessionsBundle,
...
+                {
+                    channelOutput.Source.AdvancedSequencing.CreateAdvancedSequenceStep(true);
+                    Utilities.ApplyStepProperties(channelOutput, perStepProperties[i]);
+                }
+                if (!setAsActiveSequence)
+                {
+                    channelOutput.Source.AdvancedSequencing.ActiveAdvancedSequence = string.Empty;
+                }
+            }
+            catch (Exception ex) when (ex is Ivi.Driver.OperationNotSupportedException operationNotSupported && operationNotSupported.InnerException != null && operationNotSupported.InnerException is Ivi.Driver.IviCDriverException cDriverException && cDriverException.ErrorCode == AttributeIdNotRecognized)
```
lornazh: I think the check `operationNotSupported.InnerException != null` could be omitted.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -381,6 +381,192 @@ public void DifferentSMUDevices_ForceVoltageWithPerPinPerSiteSettingsObject_Corr
...
+            var sessionManager = Initialize(pinMapWithChannelGroup: false);
+            var sessionsBundle = sessionManager.DCPower("VDD");
+
+            sessionsBundle.ConfigureMeasureWhen(DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete);
+            var steps = new List<DCPowerAdvancedSequenceStepProperties>
+            {
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 1.0, OutputFunction = DCPowerSourceOutputFunction.DCVoltage },
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 2.0, ApertureTime = 0.016, OutputFunction = DCPowerSourceOutputFunction.DCVoltage },
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 3.0, SourceDelay = 0.5, OutputFunction = DCPowerSourceOutputFunction.DCVoltage }
+            };
```
lornazh: I somehow remember that all steps must configure exactly the same properties (otherwise driver will error out), did I understand it wrong?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -74,6 +80,28 @@ internal static class Utilities
...
+        /// <summary>
+        /// Caches mappings between properties of DCPowerAdvancedSequenceStepProperties and their corresponding DCPowerAdvancedSequenceProperty enum values.
+        /// </summary>
+        /// <remarks>
+        /// This cache is populated on first access and eliminates the expensive reflection operations (GetProperties() and Enum.TryParse()) from being executed repeatedly. The cache includes
+        /// properties that have a matching enum value.
+        /// Thread Safety: Uses LazyThreadSafetyMode.ExecutionAndPublication to ensure the initialization function executes only once across all threads, with all threads seeing the same cached result.
+        /// This makes it safe for parallel access without additional locking.
+        /// </remarks>
+        private static readonly Lazy<(PropertyInfo Property, DCPowerAdvancedSequenceProperty EnumValue)[]> _propertyMappingsCache =
```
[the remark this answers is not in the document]
lornazh: > The idea behind making this Lazy was not to do this mapping(as this is heavy on refection) if code does not need it

I understand that. But the customer needs to bear the slowness once when advanced step properties are used during configuring/forcing. Whereas setup time (where `SetupDCPowerInstrumentation` is called) is usually not something customers care about.

Regarding thread safety, `SetupDCPowerInstrumentation` is typically called in Process Setup in TestStand, which is implicitly hardcoded to run as one thread only, no matter how multisiting option is configured. That's why I said if the mapping happens during `SetupDCPowerInstrumentation` we don't have to deal with thread safety.

But theoretically a customer could use `SetupDCPowerInstrumentation` outside TestStand and it's possible to be invoked in multiple threads, so maybe it's beneficial to make it lazy and handle thread safety like you did.

How about we keep what you have now and make sure it's called once in `SetupDCPowerInstrumentation`?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
```diff
@@ -89,5 +117,476 @@ public static IEnumerable<TriggerType> GetUnsupportedTriggerTypes(string modelSt
                 }
             }
         }
+
+        internal static void ApplyStepProperties(
```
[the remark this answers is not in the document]
Ashutosh-NI: Updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Types.cs
```diff
@@ -187,6 +187,1113 @@ public DCPowerWaveformResults(DCPowerFetchResult result, double deltaTime)
...
+        #region Instrument Mode
+
+        /// <summary>
+        /// Specifies the mode of operation for an instrument channel for instruments that support multiple modes.
+        /// </summary>
+        public DCPowerInstrumentMode? InstrumentMode { get; set; }
+
+        #endregion
+
+        #endregion
```
lornazh: To be consistent, let's insert an empty line after `#endregion`.

BTW, is it allowed to nest `#region`s?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Types.cs
```diff
@@ -187,6 +187,1113 @@ public DCPowerWaveformResults(DCPowerFetchResult result, double deltaTime)
...
+        /// <summary>
+        /// Default constructor for DCPowerAdvancedSequenceStepProperties.
+        /// </summary>
+        public DCPowerAdvancedSequenceStepProperties()
+        {
+        }
+
+        #region Apply To Method
```
lornazh: To be consistent, let's insert an empty line after `#region`.

BTW, is the region necessary? A method could collapse itself.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Types.cs
```diff
@@ -187,6 +187,1113 @@ public DCPowerWaveformResults(DCPowerFetchResult result, double deltaTime)
...
+            if (LcrVoltageRange.HasValue)
+            {
+                channelOutput.LCR.Advanced.VoltageRange = LcrVoltageRange.Value;
+            }
+            if (InstrumentMode.HasValue)
+            {
+                channelOutput.InstrumentMode = InstrumentMode.Value;
+            }
+        }
+        #endregion
```
lornazh: To be consistent, let's insert an empty line before `#endregion`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Utilities.cs
Ashutosh-NI: Resolving this comment as the conversation about related to this already happened in [comment ](https://github.com/ni/semi-test-library-dotnet/pull/379#discussion_r2730987922)

PR #465, opened 2026-04-06
[SMU Gang Pin Group] - Add output function and voltage limits validation for ConfigureSourceSettings and ForceCurrent methods

gmrsekar:
This PR contains code changes that add validation of per pin output function and voltage limits for all the pins in a ganged pin group.

For ganged pins, output function and voltage limits should be the same for all the pins in the pin group. `ValidatePinValuesForCascading` and `ValidateOutputFunctionForCascading` methods carry out that validation and throw error when the input has different output function or voltage limits among ganged pins. 

Changed the access modifier of the `ConfigureSourceSettings` method from public to internal as this method was created for 26.0 release and should not be public.

Added the following tests and all of them are passing.
- `DifferentSMUDevicesGanged_ForceCurrentWithDifferentPerPinSettingsObject_ThrowsException`
- `DifferentSMUDevicesGanged_ConfigureDifferentPerPinCurrentSourceSettingsForSameSite_ThrowsException`
- `DifferentSMUDevicesGanged_ConfigureDifferentPerPinCurrentSourceSettings_ThrowsException`
- `DifferentSMUDevicesGanged_ConfigureDifferentOutputFunctionForPins_ThrowsException`


nitamil (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
gmrsekar: For `ForceVoltage` even if we don't pass the outputFunction, the `GetValueFromSourceSettings` method takes the otuputFunction from each of the pins sourceSettings which would be `DCVoltage` anyways. So we're only passing the outputFunction explicitly on `ForceCurrent` methods.

SemiconductorTestLibrary.Extensions/source/Properties/AssemblyInfo.cs
```diff
@@ -1,19 +1,24 @@
...
 using System.Resources;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 // General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
 [assembly: AssemblyTitle("NationalInstruments.SemiconductorTestLibrary.Extentions")]
-[assembly: AssemblyDescription("NI Semiconductor Test Library Abstractions")]
+[assembly: AssemblyDescription("NI Semiconductor Test Library Extentions")]
```
gmrsekar: I hope the assembly name change is needed.

SemiconductorTestLibrary.Extensions/source/Properties/AssemblyInfo.cs
```diff
@@ -1,19 +1,24 @@
...
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("National Instruments")]
 [assembly: AssemblyProduct("NationalInstruments.SemiconductorTestLibrary")]
 [assembly: AssemblyCopyright("Copyright (C) National Instruments 2024. MIT License.")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 
+// This attribute makes the internal types of this assembly visible to the the unit test project
```
gmrsekar, line 18: Added this attribute since `ConfigureSourceSettins` overload was changed from public to internal and is needed in tests.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
lornazh: Oops, I think I put the comment on the wrong methods. It should be put on the `ForceVoltage` methods that take `IDictionary<string, DCPowerSourceSettings>` and `PinSiteData<DCPowerSourceSettings>`.

For those methods, what if the `DCPowerSourceSettings.OutputFunction` is incorrectly set to `DCCurrent`? We will silently correct this mistake later, but the validation logic needs a correct `OutputFunction` setting to do the right thing.

Does that make sense?

SemiconductorTestLibrary.Extensions/source/Properties/AssemblyInfo.cs
lornazh: I believe it is, it's just weird to be included in this PR... Maybe @Utkarsh-NI-Emerson could raise a separate PR for this change?

gmrsekar: Sure, I'll revert the change.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
gmrsekar: In that scenario, yes we need to pass outputFunction to do the right validation. I'll update the code.

SemiconductorTestLibrary.Extensions/source/Properties/AssemblyInfo.cs
lornazh: @gmrsekar Please make sure @Utkarsh-NI-Emerson is aware, thanks.

lornazh (approved):

SemiconductorTestLibrary.Extensions/source/Properties/AssemblyInfo.cs
Utkarsh-NI-Emerson: @lornazh 
I have created PR: https://github.com/ni/semi-test-library-dotnet/pull/478, for this change.

cc: @gmrsekar

PR #469, opened 2026-04-08
[SMU Gang Pin Group] - Throw exception for Advanced and Synchronized Sequence With Ganging

nitamil:

This PR includes logic to check for synchornized HLS on Ganged pingroup and throws exception that it is not supported yet.

We are only supporting ganging for single set point operation and normal ForceVoltageSequence, ForceCurrentSequence. We are not supporting SequenceSynchronized and AdvanceSequenceSynchronized operations in the current release. so we need to throw exception if user tries to program for the same.

Create the following auto tests for the validation and they are all passing. 

- `DifferentSMUDevicesGangedAndConfigureAdvanceSequence_ConfigureAndInitiateAdvancedSequence_ThrowsException`
- `DifferentSMUDevicesGanged_ForceVoltageSequenceSynchronized_ThrowsException`
- `DifferentSMUDevicesGanged_ForceCurrentSequenceSynchronized_ThrowsException`
- `DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedWithPerSiteValues_ThrowsException`
- `DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedWithPerPinPerSiteValues_ThrowsException`
- `DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedAndFetch_ThrowsException`
- `DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedAndFetchWithPerSiteSequenceThrowsException`
- `DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedAndFetchWithPerPinPerSiteSequenceThrowsException`


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -76,19 +78,18 @@ public static void Initiate(this DCPowerSessionsBundle sessionsBundle)
         /// <param name="sequenceTimeoutInSeconds">The maximum time, in seconds, to wait for the sequence to complete. Used only if <paramref name="waitForSequenceCompletion"/> is <see langword="true"/>. Must be greater then zero.</param>
         public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsBundle, string sequenceName, bool waitForSequenceCompletion = false, double sequenceTimeoutInSeconds = 5.0)
```
lornazh, line 98: I believe we expect this method to be called from basic HLS unsynchronized case as well.

@Ashutosh-NI can you confirm?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -336,6 +337,10 @@ public static PinSiteData<DCPowerWaveformResults> FinishWaveformAcquisition(this
         /// <returns>The per-site per-pin waveform results.</returns>
         public static PinSiteData<DCPowerFetchResult> AcquireSynchronizedWaveforms(this DCPowerSessionsBundle sessionsBundle, double apertureTimeInSeconds = 0, double measurementTimeInSeconds = 0)
         {
+            if (sessionsBundle.HasGangedChannels)
+            {
+                throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.DCPower_GangedPinGroupDetected));
+            }
```
lornazh: Make it a helper method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh, line 1968: Technically we could create a `ConfigureAdvancedSequenceCore` that extends on sessions bundle object, so that to remove more duplicated code within the 3 `ConfigureAdvancedSequence` overloads, just like how `ForceAdvancedSequenceSynchronizedCore` works.

We could make use of the existing `SequenceProvider` delegate (the naming might not be very appropriate though, a rename might be needed), or add another `StepPropertyProvider` delegate. Either way, we might want to update `ForceAdvancedSequenceSynchronized` and `ForceAdvancedSequenceSynchronizedAndFetch` overloads too, because they are actually also getting properties instead of sequences.

CC: @Ashutosh-NI

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -989,6 +1006,325 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+                    new DCPowerSourceSettings
+                    {
+                        OutputFunction = DCPowerSourceOutputFunction.DCVoltage,
+                        LimitSymmetry = DCPowerComplianceLimitSymmetry.Symmetric,
+                        Level = 2.5,
+                        Limit = 0.1
+                    }
+                }
+            });
+            void ForceAdvancedSequenceSynchronizedTest() => sessionsBundle.ForceAdvancedSequenceSynchronized(sequence, sequenceLoopCount: 1, waitForSequenceCompletion: true, sequenceTimeoutInSeconds: 5.0);
```
lornazh, line 1028: Not sure if the `sequence` setup is necessary. We could simply pass a `sequence: null` to the function under test, right?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
Ashutosh-NI: Yes, this can be called for any of the methods which uses advanced sequences.(For e.g. if user calls `ConfigureVoltageSequence` with `setAsActivesSequence` as `False` then they have to call this method to inititate advanced sequence)

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Passing `null` to sequence is creating ambiguity between the different overloads we have, as the sequence is identified based on the type (PinSiteData or SiteData or simple Double array) and corresponding overload is called.

I have reduced the size of the inputs by having just one DCPowerSourceSettings object per array.

lornazh: Oh I guess you're right.

> I have reduced the size of the inputs by having just one DCPowerSourceSettings object per array.

I think you could even instantiate a `DCPowerSourceSettings` without initialization, i.e. `new DCPowerSourceSettings()`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
gmrsekar: I was thinking about not throwing exception in this method, since for advanced sequence we are anyways throwing exceptions in `ConfigureAdvancedSequence` method and all `ForceAdvancedSequenceSynchronized` methods.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
gmrsekar: Added helper method and using it for throwing exception.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
lornazh: My point was, since it could be used for basic HLS unsynchronized case that we support, it should not throw.

gmrsekar: Yes, removed the exception throwing part.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Updated the tests.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Am creating a separate PR to refactor this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -3085,6 +3060,14 @@ private static void ReleaseSynchronizedAdvancedSequenceResources(this DCPowerSes
             // The start trigger must be set to None before any subsequent SinglePoint operations can be performed.
             sessionsBundle.DisableTriggers(new[] { TriggerType.StartTrigger });
         }
+
+        internal static void ThrowExceptionForGangedPinGroups(this DCPowerSessionsBundle sessionsBundle)
```
lornazh: Let's rename to `ValidateNoChannelGanged`.

gmrsekar: Updated the method name.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,175 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+            {
+                new[]
+                {
+                    new DCPowerSourceSettings()
+                },
+                new[]
+                {
+                    new DCPowerSourceSettings()
+                }
+            });
```
lornazh: Would the following work?
```
var sequence = new SiteData<DCPowerSourceSettings[]>(Array.Empty<DCPowerSourceSettings[]>());
```

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,175 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+            void ForceAdvancedSequenceSynchronizedTest() => sessionsBundle.ForceAdvancedSequenceSynchronized(sequence, sequenceLoopCount: 1, waitForSequenceCompletion: true, sequenceTimeoutInSeconds: 5.0);
+
+            var exception = Assert.Throws<NISemiconductorTestException>(ForceAdvancedSequenceSynchronizedTest);
+            var exceptionMessage = "This feature is not supported on a ganged pin group";
+            Assert.Contains(exceptionMessage, exception.Message);
+            sessionsBundle.ClearActiveAdvancedSequence();
```
lornazh: I guess we want to call `UngangPinGroup` before exiting the test? And since `ForceAdvancedSequenceSynchronized` throws directly before any real logic can execute, I assume we don't have to call `ClearActiveAdvancedSequence`?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,175 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+                new DCPowerSourceSettings()
+            };
+            var sequence = new PinSiteData<DCPowerSourceSettings[]>(
+                new[] { "VCC1", "VCC2", "VCC3" },
+                new[]
+                {
+                    new SiteData<DCPowerSourceSettings[]>(sites, vcc1Sequence),
+                    new SiteData<DCPowerSourceSettings[]>(sites, vcc2Sequence),
+                    new SiteData<DCPowerSourceSettings[]>(sites, vcc3Sequence)
+                });
```
lornazh: Would the following work?
```
var sequence = new PinSiteData<DCPowerSourceSettings[]>(Array.Empty<string>(), Array.Empty<int>(), data: null);
```

gmrsekar: Yes, it does. Updated the code.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: This works.. Updated the code.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Yeah, since we're not configuring the sequence, we don't have to clear it. Calling `UngangPinGroup` in all the tests now.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,114 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+        {
+            var sessionManager = Initialize("SMUGangPinGroup_SessionPerChannel.pinmap");
+            var sessionsBundle = sessionManager.DCPower(AllPinsGangedGroup);
+            sessionsBundle.GangPinGroup(AllPinsGangedGroup);
+
+            CreateDCPowerAdvancedSequencePropertyMappingsCache();
+            var sequence = new[]
+            {
+                new DCPowerSourceSettings()
+            };
```
lornazh: I believe you could create an empty array here too using `Array.Empty<T>()`.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,114 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+                measurementTimeoutInSeconds: 10.0);
+
+            var exception = Assert.Throws<NISemiconductorTestException>(ForceAdvancedSequenceSynchronizedAndFetchTest);
+            var exceptionMessage = "This feature is not supported on a ganged pin group";
+            Assert.Contains(exceptionMessage, exception.Message);
+            sessionsBundle.UngangPinGroup(AllPinsGangedGroup);
+        }
+
+        [Fact]
+        public void DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedAndFetchWithPerSiteSequenceThrowsException()
```
lornazh: Missing an underscore before "ThrowsException" in test name.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -999,6 +1016,114 @@ public void DifferentSMUDevices_ForceAdvancedSequenceSynchronizedAndFetchWithPer
...
+                measurementTimeoutInSeconds: 10.0);
+
+            var exception = Assert.Throws<NISemiconductorTestException>(ForceAdvancedSequenceSynchronizedAndFetchTest);
+            var exceptionMessage = "This feature is not supported on a ganged pin group";
+            Assert.Contains(exceptionMessage, exception.Message);
+            sessionsBundle.UngangPinGroup(AllPinsGangedGroup);
+        }
+
+        [Fact]
+        public void DifferentSMUDevicesGanged_ForceAdvancedSequenceSynchronizedAndFetchWithPerPinPerSiteSequenceThrowsException()
```
lornazh: Missing an underscore before "ThrowsException" in test name.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -1189,6 +1314,27 @@ public void DifferentSMUDevices_ConfigureAdvancedSequenceWithNotSupportedPropert
...
+        [Fact]
+        public void DifferentSMUDevicesGanged_ConfigureAdvancedSequence_ThrowsException()
+        {
+            var sessionManager = Initialize("SMUGangPinGroup_SessionPerChannel.pinmap");
+            var sessionsBundle = sessionManager.DCPower("AllPinsGangedGroup");
+            sessionsBundle.GangPinGroup("AllPinsGangedGroup");
+
+            CreateDCPowerAdvancedSequencePropertyMappingsCache();
+            string sequenceName = "Sequence";
+            var stepProperties = new List<DCPowerAdvancedSequenceStepProperties>
```
lornazh: I guess we could create an empty array and convert it to `List`.
```
var stepProperties = Array.Empty<DCPowerAdvancedSequenceStepProperties>().ToList();
```

lornazh (approved):
Except for the request to fix 2 of the test names, others are just suggestions to simplify code, you can decide whether to take or drop them.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Created an empty array.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Added the missing underscore.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Added the missing underscore.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
gmrsekar: Created an empty array of type `DCPowerAdvancedSequenceStepProperties` and converting it to `List`.

gmrsekar:
> Except for the request to fix 2 of the test names, others are just suggestions to simplify code, you can decide whether to take or drop them.

It is a good idea to pass empty array in all tests since we don't care about the input sequence or step properties. Updated all the tests to pass empty array.

gmrsekar (approved):

PR #472, opened 2026-04-09
Enable sequence mode and wait for sequence completion in InitiateAdvancedSequence

Ashutosh-NI:
This pull request updates the `InitiateAdvancedSequence` method in the DCPower control abstraction to improve advanced sequence initiation and event handling. The changes ensure the correct source mode is set and the appropriate event is used for waiting on sequence completion.

**Improvements to advanced sequence initiation:**

* The `Source.Mode` property is now explicitly set to `DCPowerSourceMode.Sequence` before initiating the advanced sequence, ensuring the hardware is in the correct mode before initiating the advanced sequence, as there might be scenario where user configure the advanced sequences in a step and in between user had a operation which changes Source mode to single point, so before initiating we have make the source mode to sequence .

**Event handling update:**
* The method now waits for the `SequenceEngineDoneEvent` instead of the `SourceCompleteEvent` to more accurately detect when the sequence has finished executing.

## Testing
Following is the snap shot of the validation on hardware where sequence loop count was 4 and wait time was 15.
and one set of sequence was value of `1.5, 2.1` for each of the pins
image

>[!NOTE]
> `InitiateAdvancedSequence` is now calling `sessionsBundle.Initiate()` as then it will be ganging compliant (which will come as a separate PR from @nitamil )


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -78,14 +79,25 @@ public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsB
...
+            sessionsBundle.Initiate();
+
+            if (waitForSequenceCompletion)
+            {
+                sessionsBundle.Do(sessionInfo =>
+                {
+                    {
+                        sessionInfo.AllChannelsOutput.Events.SequenceEngineDoneEvent.WaitForEvent(PrecisionTimeSpan.FromSeconds(sequenceTimeoutInSeconds));
+                    }
+                });
```
lornazh: Since you've already breaking a single parallel loop into 3 parallel loops sequentially, I guess you don't mind replacing this hand write extension method with the existing one we expose `WaitForEvent` from TriggersAndEvents.cs?

lornazh:
Why isn't there an auto test for this change?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
Ashutosh-NI: Added `WaitForEvent` (forgot about this), thanks!

Ashutosh-NI:
> Why isn't there an auto test for this change?

Added a test to have a single point operation before initiating, and for the `SequenceEngineDoneEvent`, I am not sure we can validate it with auto test, so adding a snapshot of which was ran on hardware.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -78,14 +79,17 @@ public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsB
...
-                allChannelOutput.Control.Initiate();
-                if (waitForSequenceCompletion)
-                {
-                    allChannelOutput.Events.SourceCompleteEvent.WaitForEvent(PrecisionTimeSpan.FromSeconds(sequenceTimeoutInSeconds));
-                }
             });
+
+            sessionsBundle.Initiate();
+
+            sessionsBundle.WaitForEvent(EventType.SequenceEngineDoneEvent, sequenceTimeoutInSeconds);
```
Mattjet27: This should still be gated by the `waitForSequenceCompletion` flag.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
```diff
@@ -121,5 +121,31 @@ public void DifferentSMUDevicesAndConfigureAdvanceSequence_InitiateClearedAdvanc
...
+            var stepProperties = new List<DCPowerAdvancedSequenceStepProperties>
+            {
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 1.0, OutputFunction = DCPowerSourceOutputFunction.DCVoltage },
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 2.0, OutputFunction = DCPowerSourceOutputFunction.DCVoltage },
+                new DCPowerAdvancedSequenceStepProperties { VoltageLevel = 3.0, OutputFunction = DCPowerSourceOutputFunction.DCVoltage }
+            };
+            sessionsBundle.ConfigureAdvancedSequence(sequenceName, stepProperties, setAsActiveSequence: false);
+
+            sessionsBundle.ForceVoltage(1.5);
+            sessionsBundle.InitiateAdvancedSequence(sequenceName);
```
Mattjet27: `waitForSequenceCompletion` is set to `false` by default, please use a named parameter argument for this to explicitly set this to `true`. 

Also, we may want to add the hardware trait attributes to ensure gets tested on hardware to confirm it runs without encounter a timeout.

Mattjet27 (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
Ashutosh-NI: Updated to be gated on flag

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
Ashutosh-NI: Added waitForSequenceCompletion

Mattjet27 (approved):

lornazh (approved):

PR #480, opened 2026-04-10
Add poc for SPI protocol

Shalini-Subramanian:

PoC for Example using SPI

I have used the SPI Register Map template provided by Digital pattern editor
image

-

-


Examples/source/Sequence/SPI/Code Modules/source/TestSteps/TestStep.cs
```diff
@@ -0,0 +1,57 @@
...
+            ISemiconductorModuleContext semiconductorModuleContext,
+            string[] spiPins,
+            byte registerAddress,
+            byte data,
+            string publishedDataId)
+        {
+            TSMSessionManager sessionManager = new TSMSessionManager(semiconductorModuleContext);
+            DigitalSessionsBundle spi = sessionManager.Digital(spiPins);
+
+            // Write: encode address + data into source waveform and burst the write pattern.
```
ni-ramya: There could be another helper method called 'WriteRegister', which would take in the digitalSessionBundle, registeraddress and data. 

Similar one for 'ReadRegister', which would take in the digitalSessionBundle, registeraddress and would return the data

Examples/source/Sequence/SPI/Code Modules/source/TestSteps/TestStep.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// </summary>
+        /// <param name="semiconductorModuleContext">The <see cref="ISemiconductorModuleContext"/> object.</param>
+        /// <param name="spiPins">The SPI port pin names.</param>
+        /// <param name="registerAddress">The 7-bit register address to write and read.</param>
+        /// <param name="data">The 8-bit data value to write.</param>
+        /// <param name="publishedDataId">The published data identifier for the read-back result.</param>
+        /// <returns>The per-site decoded register values read back from the DUT.</returns>
+        public static SiteData<byte> WriteAndReadRegister(
+            ISemiconductorModuleContext semiconductorModuleContext,
+            string[] spiPins,
```
ni-ramya: This could be made as a pingroup as discussed offline

Examples/source/Sequence/SPI/Code Modules/source/TestSteps/TestStep.cs
Shalini-Subramanian: Initially I thought of it, but the issue with that flow is
- When WriteRegister is called `TSMSessionManager sessionManager = new TSMSessionManager(semiconductorModuleContext);` is registered once.
- When ReadRegister is called, again a new tsm manager should be initialized. Because of this, both the sessions becomes independant and doesnot rely on the one another.

To avoid this, I have made this user flow altogether inside a single method

Shalini-Subramanian:
Maked as abandoned

PR #481, opened 2026-04-12
Bug Fix for `InitiateChannels`.

nitamil:

This PR contains the Bug fix for SMU Hardware Level sequencing for `FV and FI sequences`

This functionality of the HLS is valid without this bug fix when user waits for the sequence to conplete.

We found this bug during Hardware level validation and the fix is also validated immediately in the hardware.

image

Added additional tests 
1) `GangedSMUBundle_ForceVoltageWithoutWaitForSourceCompletion_ReturnsImmediately`
2) `GangedSMUBundle_ForceVoltageAndWaitForSourceCompletion_WaitsForSourceDelayToComplete`
3) `GangedSMUBundle_ForceVoltageSequenceWithoutWaitForSequenceCompletion_ReturnsImmediately`
4) `GangedSMUBundle_ForceVoltageSequenceAndWaitForSequenceCompletion_WaitsForSequenceToComplete`

Tested all four tests on Hardware and they are all passing. 
image


lornazh (changes requested):
> No addtional Auto test is need to validate the same. We found this bug during Hardware level validation and the fix is also validated immeidately in the hardware.

I don't think this is valid reasoning. The purpose of adding tests is not just to verify your current changes, is more to protect future changes to break expected behavior.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Should we change the parameter name from `waitForSourceCompletion` to something like `waitForSourceOrSequenceCompletion`, since this method now handles the sequencing part as well?

lornazh: Good point! Or even simpler, `waitForCompletion` indicating waiting for initiate to complete?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -39,6 +39,16 @@ public TSMSessionManager Initialize(string pinMapFileName)
             return new TSMSessionManager(_tsmContext);
         }
 
+        public DCPowerSessionsBundle GetGangedSMUBundle()
```
lornazh: Please make it private helper.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -4846,6 +4856,92 @@ public void DifferentSMUDevices_ConfigureDifferentSourceSettings_CorrectValuesAr
...
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        public void GangedSMUBundle_ForceVoltageAndWaitForSourceCompletion_WaitsForSourceDelayToComplete()
+        {
+            var sessionsBundle = GetGangedSMUBundle();
+            sessionsBundle.ConfigureSourceDelay(0.1);
+            var startTime = DateTime.Now;
+
+            sessionsBundle.ForceVoltage(voltageLevel: 3.3, currentLimit: 0.1, waitForSourceCompletion: true);
+            var elapsedTime = (DateTime.Now - startTime).TotalSeconds;
```
lornazh: Can we also use `TotalMilliseconds` to be consistent?

lornazh (approved):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: Changed the scope to private.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: Made all elapsed time to `TotalMilliSeconds`

PR #484, opened 2026-04-15
Refactoring Sequence and Ganged Sequence methods

gmrsekar:

This PR contains code refactoring of `ConfigureAdvanceSequence`, `ConfigureVoltageSequence`, `ConfigureCurrentSequence`, `ConfigureVoltageSequenceWithSourceDelays` and `ConfigureCurrentSequenceWithSourceDelays` methods and their overloads.

This is a general code refactoring of the methods that are part of HLS and Ganging + HLS features.

No new test has been added since no new functionality is added. All the existing tests are passing.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1962,16 +1941,8 @@ public static void ConfigureAdvancedSequence(
             bool commitFirstElementAsInitialState = false)
         {
             sessionsBundle.ValidateNoChannelGanged();
```
lornazh: I believe this check should be part of the new `ConfigureAdvancedSequenceCore` method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -26,6 +26,22 @@ public static class Source
         /// <returns>An array of type T containing the generated sequence based on the provided site pin information.</returns>
         private delegate T[] SequenceProvider<T>(SitePinInfo sitePinInfo);
 
+        /// <summary>
+        /// Provides an array of source delays of type T based on the specified site pin information.
+        /// </summary>
+        /// <typeparam name="T">The type of items returned in the array.</typeparam>
+        /// <param name="sitePinInfo">The site pin information used to generate the sequence.</param>
+        /// <returns>An array of type T containing the generated source delays based on the provided site pin information.</returns>
+        private delegate T[] SoureDelayProvider<T>(SitePinInfo sitePinInfo);
```
lornazh: How about we don't make these delegate property specific, but type specific?

In other words, I propose we have `ArrayProvider`, `IEnumerableProvider` and `ValueProvider` (or `DoubleProvider`).

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1075,21 +1091,21 @@ private static void ForceSequenceSynchronizedCore(
         /// This method will set the Source Mode back to SinglePoint mode upon returning.
         /// </remarks>
         /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
-        /// <param name="sequence">The sequence of source settings to apply.</param>
```
lornazh, line 1078: Now I read the explanation of the parameter, "sequence" here does not mean the value sequence, but settings sequence, I feel it does make sense. So please revert this parameter name change at all similar places.

Sorry for the inconvenience this has caused.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1814,43 +1824,31 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
-                sessionInfo.ConfigureSequenceForCascadingCore(
-                    sequenceName: sequenceName,
-                    sequence: sequence.GetValue(sitePinInfo.SiteNumber),
-                    sequenceLoopCount: sequenceLoopCount,
-                    outputFunction: DCPowerSourceOutputFunction.DCVoltage,
-                    sitePinInfo: sitePinInfo,
-                    sequenceStepDeltaTimeInSeconds: sequenceStepDeltaTimeInSeconds,
-                    setAsActiveSequence: setAsActiveSequence);
-            });
+            SequenceProvider<double> getSequence = (sitePinInfo) => sequence.GetValue(sitePinInfo.SiteNumber);
```
lornazh: Brackets around `sitePinInfo` is not necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1814,43 +1824,31 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
-                sessionInfo.ConfigureSequenceForCascadingCore(
-                    sequenceName: sequenceName,
-                    sequence: sequence.GetValue(sitePinInfo),
-                    sequenceLoopCount: sequenceLoopCount,
-                    outputFunction: DCPowerSourceOutputFunction.DCVoltage,
-                    sitePinInfo: sitePinInfo,
-                    sequenceStepDeltaTimeInSeconds,
-                    setAsActiveSequence: setAsActiveSequence);
-            });
+            SequenceProvider<double> getSequence = (sitePinInfo) => sequence.GetValue(sitePinInfo);
```
lornazh: Brackets around `sitePinInfo` is not necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1778,21 +1794,15 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
-                        outputFunction: DCPowerSourceOutputFunction.DCVoltage,
-                        sitePinInfo: sitePinInfo,
-                        sequenceStepDeltaTimeInSeconds,
-                        setAsActiveSequence: setAsActiveSequence);
-                });
+                SequenceProvider<double> getSequence = _ => sequence;
+                sessionsBundle.ConfigureSequenceForCascadingCore(
+                    hasGangedChannels: true,
+                    sequenceName: sequenceName,
+                    getSequence: getSequence,
```
lornazh: I feel we could directly define the provider delegate here for simplicity.
```
getSequence: _ => sequence,
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1905,42 +1897,29 @@ public static void ConfigureCurrentSequence(this DCPowerSessionsBundle sessionsB
...
         {
-            sessionsBundle.ValidatePinsForGanging(sessionsBundle.HasGangedChannels);
-            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
-            {
-                var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
-                channelOutput.Control.Abort();
-                channelOutput.Source.Output.Function = DCPowerSourceOutputFunction.DCCurrent;
-                sessionInfo.ConfigureSequenceForCascadingCore(
-                    sequenceName: sequenceName,
-                    sequence: sequence.GetValue(sitePinInfo, out bool isGroupData),
```
lornazh, line 1936: Your refactor has a behavior change in this case, you're ignoring the `needDataAdjustment` assignment, is it why some tests are failing?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2833,6 +2769,58 @@ private static void ConfigureAdvancedSequenceCore(
...
+                var sourceDelay = getSourceDelays?.Invoke(sitePinInfo);
+                sessionInfo.ConfigureSequenceForCascadingCore(
+                    sequenceName: sequenceName,
+                    sequence: sequence,
+                    sequenceLoopCount: sequenceLoopCount,
+                    outputFunction: outputFunction,
+                    sitePinInfo: sitePinInfo,
+                    sequenceStepDeltaTimeInSeconds: sequenceStepDeltaTimeInSeconds,
+                    sourceDelay: sourceDelay,
+                    setAsActiveSequence: setAsActiveSequence);
```
lornazh: I feel it's simpler to expand the implementation of this method here. Or even update existing `ConfigureSequenceForCascadingCore` that extends on `DCPowerSessionInformation` to `DCPowerSessionsBundle`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2375,8 +2309,10 @@ public static void ConfigureCurrentSequenceWithSourceDelays(
             int sequenceLoopCount = 1,
             bool setAsActiveSequence = false)
         {
-            sessionsBundle.ValidatePinsForGanging(sessionsBundle.HasGangedChannels);
-            sessionsBundle.ValidatePinValuesForCascading(sessionsBundle.HasGangedChannels, sourceDelaysInSeconds);
+            var hasGangedChannels = sessionsBundle.HasGangedChannels;
+            sessionsBundle.ValidatePinsForGanging(hasGangedChannels);
+            sessionsBundle.ValidatePinValuesForCascading(hasGangedChannels, sequence);
```
lornazh: Why are we adding this check? Is it intentional? Is it a bug earlier?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2833,6 +2769,58 @@ private static void ConfigureAdvancedSequenceCore(
             }
         }
 
+        private static void ConfigureAdvancedSequenceCore<T>(
+            this DCPowerSessionsBundle sessionsBundle,
+            string sequenceName,
+            StepPropertyProvider<T> getPerStepProperties,
```
lornazh: Can this be `Provider<DCPowerAdvancedSequenceStepProperties>`? Then I feel we don't need the cast on line 2782?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2833,6 +2769,58 @@ private static void ConfigureAdvancedSequenceCore(
...
+            {
+                var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
+                var perStepProperties = getPerStepProperties(sitePinInfo).Cast<DCPowerAdvancedSequenceStepProperties>();
+                channelOutput.ConfigureAdvancedSequenceCore(
+                    sequenceName,
+                    sitePinInfo.ModelString,
+                    perStepProperties,
+                    setAsActiveSequence: setAsActiveSequence,
+                    commitFirstElementAsInitialState: commitFirstElementAsInitialState);
+                channelOutput.Source.Mode = DCPowerSourceMode.Sequence;
```
lornazh: Why is this call? Isn't it be called in `channelOutput.ConfigureAdvancedSequenceCore` already?

lornazh (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1778,21 +1794,15 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
-                    channelOutput.Source.Output.Function = DCPowerSourceOutputFunction.DCVoltage;
-                    sessionInfo.ConfigureSequenceForCascadingCore(
-                        sequenceName: sequenceName,
-                        sequence: sequence,
-                        sequenceLoopCount: sequenceLoopCount,
-                        outputFunction: DCPowerSourceOutputFunction.DCVoltage,
-                        sitePinInfo: sitePinInfo,
-                        sequenceStepDeltaTimeInSeconds,
-                        setAsActiveSequence: setAsActiveSequence);
-                });
```
lornazh, line 1795: I actually was not asking for a refactor for this logic, what I asked was just `ConfigureAdvancedSequenceCore`.

So for this one, maybe we just move the 3 lines outside the existing `ConfigureSequenceForCascadingCore` into it?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Moved the check inside `ConfigureAdvancedSequenceCore` method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Yeah, that seems better. Updated the delegate names.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Test failures were due to passing incorrect values to parameters in one Configure method. Corrected that and now all tests are passing. As for the `needDataAdjustment` parameter, I noticed it in `ConfigureCurrentSequenceWithSourceDelays` method and decided to let it call `sessionInfo.ConfigureSequenceForCascadingCore` for passing this parameter. Missed it here. Making the change now.

Just realized that the auto tests are not asserting the sequence set, if tests are running in offline mode. That's why this change didn't cause any test failure.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Removed the brackets.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: This check should happen for sequence as well. So, ideally it should have been there already.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the code.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Unintended change. Removed it.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: I thought having the same code as in `sessionInfo.ConfigureSequenceForCascadingCore` would be duplicate so called this method here. One overload of `ConfigureSequenceForCascadingCore ` needs to be extending on `SessionInfo` as it is used in other places where we cannot call it on `SessionsBundle`. So, I'd rather expand the implementation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Removed the brackets.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Moved the abort call inside the method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the code in all similar places.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: No issues. Reverted the change.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1113,10 +1121,10 @@ public static void ForceAdvancedSequenceSynchronized(
             bool waitForSequenceCompletion = false,
             double sequenceTimeoutInSeconds = 5.0)
         {
-            SequenceProvider<DCPowerSourceSettings> getVoltageSequence = sitePinInfo => sequence.GetValue(sitePinInfo.SiteNumber);
+            IEnumerableProvider<DCPowerSourceSettings> getSequence = sitePinInfo => sequence.GetValue(sitePinInfo.SiteNumber);
```
gmrsekar: Saw `getVoltageSequence ` only in this line. Hope it is fine to change this also to `getSequence` to be consistent.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh: “Provides **an array of** items of type T ..."?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -24,7 +24,15 @@ public static class Source
         /// <typeparam name="T">The type of items returned in the sequence.</typeparam>
         /// <param name="sitePinInfo">The site pin information used to generate the sequence.</param>
         /// <returns>An array of type T containing the generated sequence based on the provided site pin information.</returns>
-        private delegate T[] SequenceProvider<T>(SitePinInfo sitePinInfo);
+        private delegate T[] ArrayProvider<T>(SitePinInfo sitePinInfo);
+
+        /// <summary>
+        /// Provides an array of step properties of type T based on the specified site pin information.
```
lornazh: “Provides **an IEnumerable of** items of type T ..."?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1394,7 +1402,7 @@ public static PinSiteData<SingleDCPowerFetchResult[]> ForceAdvancedSequenceSynch
 
         private static PinSiteData<SingleDCPowerFetchResult[]> ForceAdvancedSequenceSynchronizedCore<T>(
             this DCPowerSessionsBundle sessionsBundle,
-            SequenceProvider<T> getSequence,
+            IEnumerableProvider<T> getProperty,
```
lornazh: Shouldn't all `SequenceProvider` be replaced by `ArrayProvider`?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1814,43 +1814,31 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
-                    outputFunction: DCPowerSourceOutputFunction.DCVoltage,
-                    sitePinInfo: sitePinInfo,
-                    sequenceStepDeltaTimeInSeconds: sequenceStepDeltaTimeInSeconds,
-                    setAsActiveSequence: setAsActiveSequence);
-            });
+            ArrayProvider<double> getSequence = sitePinInfo => sequence.GetValue(sitePinInfo.SiteNumber);
+            sessionsBundle.ConfigureSequenceForCascadingCore(
+                hasGangedChannels: sessionsBundle.HasGangedChannels,
+                sequenceName: sequenceName,
+                getSequence: getSequence,
```
lornazh: Same suggestion, defining `getSequence` beforehand does not seem necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh: I meant moving the `Abort` and the set of `Output.Function` into existing `ConfigureSequenceForCascadingCore` and not creating a new overload on sessions bundle anymore.

Which, will lead this block of code look like
```
                sessionsBundle.ValidatePinsForGanging(hasGangedChannels: true);
                sessionsBundle.Do((sessionInfo, sitePinInfo) =>
                {
                    sessionInfo.ConfigureSequenceForCascadingCore(
                        sequenceName: sequenceName,
                        sequence: sequence,
                        sequenceLoopCount: sequenceLoopCount,
                        outputFunction: DCPowerSourceOutputFunction.DCVoltage,
                        sitePinInfo: sitePinInfo,
                        sequenceStepDeltaTimeInSeconds,
                        setAsActiveSequence: setAsActiveSequence);
                });
```

Does this make sense?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1778,28 +1785,20 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
...
+                    outputFunction: DCPowerSourceOutputFunction.DCVoltage,
+                    sequenceStepDeltaTimeInSeconds: sequenceStepDeltaTimeInSeconds,
+                    setAsActiveSequence: setAsActiveSequence);
             }
             else
             {
                 sessionsBundle.Do(sessionInfo =>
                 {
                     sessionInfo.Session.Control.Abort();
-                    sessionInfo.AllChannelsOutput.Source.Output.Function = DCPowerSourceOutputFunction.DCVoltage;
```
lornazh, line 1802: Why is this line removed?

Let's don't add or remove method calls so as to be safe.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1869,28 +1856,20 @@ public static void ConfigureCurrentSequence(this DCPowerSessionsBundle sessionsB
...
+                    outputFunction: DCPowerSourceOutputFunction.DCCurrent,
+                    sequenceStepDeltaTimeInSeconds: sequenceStepDeltaTimeInSeconds,
+                    setAsActiveSequence: setAsActiveSequence);
             }
             else
             {
                 sessionsBundle.Do(sessionInfo =>
                 {
                     sessionInfo.Session.Control.Abort();
-                    sessionInfo.AllChannelsOutput.Source.Output.Function = DCPowerSourceOutputFunction.DCCurrent;
```
lornazh, line 1893: Same thing, why is this call removed?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: It felt redundant as we're also passing `outputFunction: DCPowerSourceOutputFunction.DCVoltage` to the `ConfigureSequenceCore` method. Safer side, I can revert this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: It felt redundant as we're also passing outputFunction: DCPowerSourceOutputFunction.DCVoltage to the ConfigureSequenceCore method. Safer side, I can revert this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Yes, Updated the code. I initially felt I can get rid of `sessionInfo.ConfigureSequenceForCascadingCore` method fully, but that wasn't the case since in some places this method couldn't be replaced. Now it makes to not have `ConfigureSequenceForCascadingCore` extending on `SessionsBundle`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Calling `sessionInfo.ConfigureSequenceForCascadingCore` now, hence not using the delegate anymore.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh: Passing the output function value does not mean we set it to driver property in the `ConfigureSequenceCore`, right?  They are completely different.

Have you checked whether there is the driver output function set logic inside `ConfigureSequenceCore` before you remove this call?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the summary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the summary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -19,12 +19,20 @@ public static class Source
         #region Custom Delegates
 
         /// <summary>
-        /// Provides a sequence of items of type T based on the specified site pin information.
+        /// Provides an array of items of type T based on the specified site pin information.
         /// </summary>
         /// <typeparam name="T">The type of items returned in the sequence.</typeparam>
         /// <param name="sitePinInfo">The site pin information used to generate the sequence.</param>
         /// <returns>An array of type T containing the generated sequence based on the provided site pin information.</returns>
```
lornazh: Update to not be specific on any concept (e.g. sequence) but focus on the array type.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -19,12 +19,20 @@ public static class Source
...
-        private delegate T[] SequenceProvider<T>(SitePinInfo sitePinInfo);
+        private delegate T[] ArrayProvider<T>(SitePinInfo sitePinInfo);
+
+        /// <summary>
+        /// Provides an IEnumerable of items of type T based on the specified site pin information.
+        /// </summary>
+        /// <typeparam name="T">The type of items returned in the sequence.</typeparam>
+        /// <param name="sitePinInfo">The site pin information used to generate the sequence.</param>
+        /// <returns>An IEnumerable of type T containing the generated step properties based on the provided site pin information.</returns>
```
lornazh: Update to not be specific on any concept (e.g. step properties) but focus on the IEnumerable type.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: I checked the method and but mistook. We're not setting driver property for output function we're only setting `DCPowerAdvancedSequenceProperty` to VoltageLevel or CurrentLevel based on the `outputFunction` passed from here. 
I can revert all such changes or set driver output function inside `ConfigureSequenceCore` method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2845,10 +2806,11 @@ private static void ConfigureSequenceForCascadingCore(
...
             bool setAsActiveSequence = false)
         {
-            var output = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
-            ValidateChannelOutputAndSitePinInfoPair(sitePinInfo, output.Name);
+            var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
+            channelOutput.Control.Abort();
+            ValidateChannelOutputAndSitePinInfoPair(sitePinInfo, channelOutput.Name);
             sequence = DivideSequenceForCascading(outputFunction, sitePinInfo, needDataAdjustment, sequence);
-            output.ConfigureSequenceCore(
+            channelOutput.ConfigureSequenceCore(
```
lornazh: Name change not necessary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2845,10 +2806,11 @@ private static void ConfigureSequenceForCascadingCore(
             double[] sourceDelay = null,
             bool setAsActiveSequence = false)
         {
-            var output = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
-            ValidateChannelOutputAndSitePinInfoPair(sitePinInfo, output.Name);
+            var channelOutput = sessionInfo.Session.Outputs[sitePinInfo.IndividualChannelString];
+            channelOutput.Control.Abort();
```
lornazh: Do validate before aborting.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh: Correct, that's the right thing to do.

gmrsekar: Setting the `outputFunction` in `ConfigureSequenceCore` method now.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the code.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Reverted the name change.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Setting the `outputFunction` now in `ConfigureSequenceCore` method now.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the summary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Updated the summary.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Yes, it should be. Updated the delegate.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -1835,13 +1835,10 @@ public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsB
         public static void ConfigureVoltageSequence(this DCPowerSessionsBundle sessionsBundle, string sequenceName, PinSiteData<double[]> sequence, int sequenceLoopCount = 1, double? sequenceStepDeltaTimeInSeconds = null, bool setAsActiveSequence = false)
         {
             var hasGangedChannels = sessionsBundle.HasGangedChannels;
-            sessionsBundle.ValidatePinsForGanging(hasGangedChannels);
+            sessionsBundle.ValidatePinsForGanging(sessionsBundle.HasGangedChannels);
```
lornazh: Revert this change.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
lornazh: I don't think so since current sequence could be different for differnt pins, right?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -2833,6 +2773,27 @@ private static void ConfigureAdvancedSequenceCore(
             }
         }
 
+        private static void ConfigureAdvancedSequenceCore(
```
lornazh, line 2779: I feel this method is missing an `Abort` call.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Oh yes. May be that's why it wasn't added in the first place. Will revert the change.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Added the missing `Abort` call.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
gmrsekar: Reverted the change.

lornazh (approved):

PR #486, opened 2026-04-17
[SMU Gang] Measure Workflow Test Fix

nitamil:

This PR Splits the Initiate Measure Test into 3 different test scenarios with testerOnly triat.

1) No Ganging
2) Ganging with ConfigureMeasureSettings
3) Ganging with ConfigureMeasureWhen

This PR fixes the test logic for `Measure Only` workflow under different measurement configurations.

All the 3 test cases are tested on a PXI testsetup.

Configure MeasureWhen test result 👍🏽 
image

Gang and MeasureSettings test result 👍🏽 
image

Gang and MeasureWhen Timeout test result 👍🏽  
image

Test results on Cauvery are passing:
image

### DIFF
--- a/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
+++ b/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
@@ -66,7 +66,6 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
             var sessionManager = Initialize(pinmap);
             var dcPower = sessionManager.DCPower(new[] { "PowerPins" });
             dcPower.ConfigureMeasureWhen(measureWhen);
-
             if (measureWhen == DCPowerMeasurementWhen.OnMeasureTrigger)
             {
                 dcPower.ConfigureTriggerSoftwareEdge(TriggerType.MeasureTrigger);
--- a/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
+++ b/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
@@ -212,8 +212,6 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
 
         [Theory]
         [Trait(nameof(Platform), nameof(Platform.TesterOnly))]
-        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
-        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
         [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
         [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnDemand)]
         [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
--- a/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
+++ b/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
@@ -51,6 +51,33 @@ namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbst
         }
 
         [Theory]
+        [Trait(nameof(Platform), nameof(Platform.TesterOnly))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        public void ConfigureMeasureWhen_Initiate_DataMeasurementDoesNotTimeOut(string pinmap, DCPowerMeasurementWhen measureWhen)
+        {
+            var sessionManager = Initialize(pinmap);
+            var dcPower = sessionManager.DCPower(new[] { "PowerPins" });
+            dcPower.ConfigureMeasureWhen(measureWhen);
+
+            if (measureWhen == DCPowerMeasurementWhen.OnMeasureTrigger)
+            {
+                dcPower.ConfigureTriggerSoftwareEdge(TriggerType.MeasureTrigger);
+            }
+
+            dcPower.Initiate();
+
+            dcPower.MeasureVoltage();
+        }
+
+        [Theory]
         [InlineData(false)]
         [InlineData(true)]
         public void DifferentSMUDevicesAndConfigureAdvanceSequenceAsInactive_InitiateAdvancedSequences_AdvanceSequenceActivated(bool pinMapWithChannelGroup)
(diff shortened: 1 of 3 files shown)


SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -5023,7 +5023,7 @@ public void GangedSMUBundle_ForceVoltageWithoutWaitForSourceCompletion_ReturnsIm
             sessionsBundle.ConfigureSourceDelay(0.1);
             var startTime = DateTime.Now;
 
-            sessionsBundle.ForceVoltage(voltageLevel: 1.0, currentLimit: 0.1, waitForSourceCompletion: false);
+            sessionsBundle.ForceVoltage(voltageLevel: 1.0, currentLimit: 0.05, waitForSourceCompletion: false);
```
nitamil, line 5026: @lornazh and @Dilpreetsingh3211, The changes (`currentLimit = 50mA`) in SourceTest.cs file is required for the test to run on 4163 hardware (`LuanYuan` STS uses this 4162/3 hardware).

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
```diff
@@ -50,6 +50,33 @@ public void Dispose()
...
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        public void ConfigureMeasureWhen_Initiate_DataMeasurementDoesNotTimeOut(string pinmap, DCPowerMeasurementWhen measureWhen)
```
lornazh, line 64: Why is this test added?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
lornazh, line 205: Is this call necessary? Does the test fail without it?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
```diff
@@ -181,5 +211,39 @@ public void GangPinGroupAndConfigureMeasure_Initiate_DataMeasurementDoesNotTimeO
...
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnDemand)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.OnMeasureTrigger)]
+        [InlineData("Mixed Signal Tests Common Session.pinmap", DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete)]
+        public void GangPinGroupAndConfigureMeasureWhen_InitiateAndMeasure_TimeOutOccurs(string pinmap, DCPowerMeasurementWhen measureWhen)
```
lornazh: Is this really necessary?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: This is to add additional test coverage for Non-Ganged Pins the individual configuration is still valid and working. Which is needed for the original change we did for skipping of operation on ganged Pins.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: Not required, but, Just to be sure that the leader pins get the triggers but not the follower pin (to be specific) about the test failure.

gmrsekar: Just to confirm, `dcPower.ConfigureMeasureSettings(dcpowerMeasureSettings);` call sets measureWhen property for both leader and follower channels right?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
lornazh: The it seems you are testing configure methods instead of `Initiate`? If that's the case, then it should be in SourceTests.cs and TriggersAndEventsTests.cs?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
lornazh: I don't think it's necessary, please remove. At the end of the day, we won't recommend customers to do this either.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: Added as part of code path coverage, but it is not required. Currently, it takes more time in the tester as the time-out is 20 seconds. I thought of discussing with you and remove it.

Utkarsh-NI-Emerson (commented):
@nitamil 
If the changes are just for the test cases, then I suggest running the ATS with this branch in `Test` mode, with latest Bundle on Cauvery and attaching the results for the run in the Testing section of the PR Description.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: @gmrsekar, yes `measureWhen` property will be configured on both but the trigger terminal for followers will be set to `leader channel's measure terminal name\path` and configured as digital edge. Whereas For Leader channel's measurement trigger will be configured as software edge on if the `measurewhen` is set to `OnMeasureTrigger`. 

@lornazh, it is needed as the test fails without the `SoftwareEdgeTrigger` configuration. The default is digital edge, and we need to provide a trigger terminal value. Instead, I am configuring it to software edge as the `MeasureVoltageAndCurrent` method has logic to send the necessary software trigger.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: Extra blank line removed. Test is for Initiate.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
nitamil: I have removed all other STS configurations except Cauvery. Let this test be there till we resolve the measurement deadlock [tech debt](https://dev.azure.com/ni/DevCentral/_workitems/edit/3830112/). This test result is an evidence that the driver code is kind of sequencing/serializing the operations if there is a common (single) session for both leader and follower channels. 

Image

nitamil:
> @nitamil If the changes are just for the test cases, then I suggest running the ATS with this branch in `Test` mode, with latest Bundle on Cauvery and attaching the results for the run in the Testing section of the PR Description.

@Utkarsh-NI-Emerson , We found a measurement trigger related issue. This PR has a solution which is required for the tests to pass on STS hardware.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs
lornazh: I don't think this test is necessary. If you want to keep it, then
1. The test has four sections while it should be three, please fix it.
2. It's the measure that times out, so don't make `Initiate` part of the method for `Assert.Throws`.
3. Technically you're testing `MeasureVoltage`, so it should exist in MeasureTests.cs.
4. If you want to keep it as evidence, you'd better be specific on asserting the timeout duration, instead of requiring manual checking for that. It should be a test that asserts the timeout is around 20 seconds, and it should fail now and be skipped.

nitamil: Removed this test from this PR. Will create a separate PR with this test.

lornazh (approved):

PR #497, opened 2026-04-27
Fix the deadlock condition for measure and fetch

Ashutosh-NI:
This pull request refactors the `MeasureVoltageAndCurrent` method in `Measure.cs` to simplify the measurement logic. The main changes involve removing parallelization.

Measurement logic simplification:

* Removed the use of `InvokeInParallel` and `Parallel.For` in favor of straightforward sequential processing (as NIDCPower driver internally sequntialise the calls or `Measure` and `Fetch` and they use the same lock.) .
* Eliminated unnecessary locking since parallel execution was removed, further simplifying the code. 

## Testing
Ran the test in class `GangTypeCaseTests`  (108) in `STSCauvery` for atlesat 23 iterations (139 mins) and were not seeing any intermittent issues  
image

Multiple ATS runs ( Ran in dev mode as the changes are in extension, but it does not matter as the issue was reproduceable in dev mode as well)
1. https://dev.azure.com/ni/DevCentral/_build/results?buildId=19083707&view=results
2. https://dev.azure.com/ni/DevCentral/_build/results?buildId=19083871&view=logs&s=96ac2280-8cb4-5df5-99de-dd2da759617d&j=9316236d-33d1-5e32-58fc-bdb37708dc8b
3. https://dev.azure.com/ni/DevCentral/_build/results?buildId=19083886&view=results
4. https://dev.azure.com/ni/DevCentral/_build/results?buildId=19083886&view=results
5. https://dev.azure.com/ni/DevCentral/_build/results?buildId=19084864&view=results


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -597,52 +596,43 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
...
-                },
-                () =>
+            // Measurement for on demand channels.
+            if (onDemandChannelIndexes.Any())
+            {
+                // Measure all channels that are configured to measure on demand as a single driver call to optimize test time.
+                var measureResult = session.Measurement.Measure(string.Join(",", onDemandChannelStrings));
+                for (int i = 0; i < onDemandChannelIndexes.Count; i++)
                 {
-                    Parallel.For(0, channelCount, channelIndex =>
```
nitamil, line 620: I feel we can still keep the parallel for loop. That way once the driver team makes change to lock on specific channel of the session, then our code will be performant. 

@lornazh, let me know your thoughts. Should we keep it or remove it as per the current changes in the PR.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -597,52 +596,43 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
...
 
-                        switch (dcOutput.Measurement.MeasureWhen)
+            for (int channelIndex = 0; channelIndex < channelCount; channelIndex++)
+            {
+                var sitePinInfo = listOfChannelsToMeasure[channelIndex];
+                var dcOutput = session.Outputs[sitePinInfo.IndividualChannelString];
+
+                switch (dcOutput.Measurement.MeasureWhen)
+                {
+                    case DCPowerMeasurementWhen.OnMeasureTrigger:
```
lornazh, line 619: Like I said, do we want to move the sending of software trigger back in this method and make it happen always before fetching?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
lornazh: @nitamil Although what you suggested is ideal, I'm afraid it might be a little complicated (at least requires more thinking and testing) to keep the parallel loop as well as making these operations happen sequentially under certain conditions, let alone we're actually not sure it will work as is with the driver improvement.

So given where we're at in the release cycle now, I recommend us keeping things simple for now. For tracking purpose, we could create a tech debt work item that reminds us to change this logic back to parallel once the driver improvement is ready for use. We could link the feature request for driver team in the tech debt work item as well.

CC: @Ashutosh-NI

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
nitamil: Let me check the logic and update the logic here. To my understanding, there is always a dependency of fetching and software trigger. 

So both on measure and sending software trigger must execute before entering in to fetch.

lornazh:
@Ashutosh-NI @nitamil Now that we understand the root cause of the deadlock, I believe [this case](https://github.com/ni/semi-test-library-dotnet/blob/e1df0166d3581ae7fee0d15b001e1ed687c2f81f/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs#L192) on Extensions layer would also intermittently fail, right?

lornazh (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Ashutosh-NI: Created the tech debt for the same [Technical Debt 3842086](https://dev.azure.com/ni/DevCentral/_workitems/edit/3842086): Make changes to MeasureVoltageAndCurrent to make the operation in parallel

Ashutosh-NI:
> @Ashutosh-NI @nitamil Now that we understand the root cause of the deadlock, I believe [this case](https://github.com/ni/semi-test-library-dotnet/blob/e1df0166d3581ae7fee0d15b001e1ed687c2f81f/SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/ControlTests.cs#L192) on Extensions layer would also intermittently fail, right?

@lornazh Yes it should, I think we were lucky with the run of that test for not catching it, but with updated code it should not error out.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Ashutosh-NI: This will be implemented as separate PR by @nitamil, so resolving this comment.

nitamil (approved):

PR #502, opened 2026-05-04
STL | Digital | Fix MeasureAndPublishVoltage throw expcetion while publishing the data

Shalini-Subramanian:
- An Exception was thrown when `MeasureAndPublishVoltage` was invoked in Digital extension. This is because, the `Measure` method was returning the Measurements for all sites including the shadow ones. Ideally, the measurement should take place only for the primary and non-shared channels. Because of this the exception was thrown due to result count mismatch.

This PR contains the fix for the above exception.

- This PR contains the fix for the above-mentioned exception.
- This was achieved by below implementation
     - The `PinSetString` was filtered only for the primary and non-shared channels.
     - Because of this the result count mismatch was fixed.

- Added unit test to verify this behavior


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/PPMU.cs
```diff
@@ -513,8 +514,14 @@ public static void ConfigureApertureTime(this DigitalSessionsBundle sessionsBund
         /// <returns>The measurements.</returns>
         public static double[] Measure(this DigitalSessionInformation sessionInfo, MeasurementType measurementType)
         {
-            var ppmu = sessionInfo.Session.PinAndChannelMap.GetPinSet(sessionInfo.PinSetString).Ppmu;
-            return ppmu.Measure(measurementType is MeasurementType.Voltage ? PpmuMeasurementType.Voltage : PpmuMeasurementType.Current);
+            var activeSitePins = sessionInfo.AssociatedSitePinList
```
hcnagaraj-e, line 517: @Shalini-Subramanian Can you test this code with disabling sites through handler and update the behavior?

hcnagaraj-e (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/PPMU.cs
Shalini-Subramanian: I have tested it already, and it is working as expected

Shalini-Subramanian: I'll add a test if you want me to add one

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/PPMU.cs
```diff
@@ -513,8 +514,14 @@ public static void ConfigureApertureTime(this DigitalSessionsBundle sessionsBund
...
+                .Where(pin => !pin.SkipOperations)
+                .Select(pin => pin.SitePinString);
+            var filteredPinSetString = string.Join(",", activeSitePins);
+            var ppmu = sessionInfo.Session.PinAndChannelMap.GetPinSet(filteredPinSetString).Ppmu;
+            var allMeasurements = ppmu.Measure(measurementType is MeasurementType.Voltage ? PpmuMeasurementType.Voltage : PpmuMeasurementType.Current);
```
lornazh: Can return directly from here.

Shalini-Subramanian: updated

Shalini-Subramanian:
@lornazh I have validated this behavior in the PXI system, and it is work fine

lornazh (approved):

PR #516, opened 2026-06-15
NIDigital Extension Methods `Commit()`, `Initiate()`,  `ConfigurePattern()` and `GetPatternStartLabel()` methods added

nitamil:

Adds new NI-Digital `DigitalSessionsBundle` extension methods 
- To support lower-level pattern control operations 
  - Added `Commit()` and `Initiate()` methods.
- To configuration and read the pattern start label, 
  - Added `ConfigurePattern(...)` and `GetPatternStartLabel()` methods.
- Along with unit tests intended to validate the new APIs.

We are adding 4 more functionality to STL nidigital drivers to ease the workflow for customers.

Created Auto tests for all the new extension methods. Ran them on my local VM and the tests are passing.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,38 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+            });
+        }
+
+        /// <summary>
+        /// Configures the pattern start label for subsequent pattern execution.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
```
NikhilSivadas-NIB: Can we consider adding validation for startLabel and throw NISemiconductorTestException with appropriate message ? Hope we will not proceed if startLabel is invalid

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,38 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.PatternControl.StartLabel = startLabel;
+
+                if (siteNumbers != null && siteNumbers.Length > 0)
+                {
+                    string siteList = string.Join(",", siteNumbers.Select(sn => $"site{sn}"));
```
NikhilSivadas-NIB: As per spec this line is "string siteList = string.Join(",", siteNumbers);", as per implementation its "implementation uses "site {sn}" formatting." Hope this is due to expected driver format?
If yes, can we update the spec to reflect the same

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,38 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.PatternControl.StartLabel = startLabel;
+
+                if (siteNumbers != null && siteNumbers.Length > 0)
```
NikhilSivadas-NIB: Would it better to add validation for negative or duplicate siteNumbers to prevent invalid values from being passed to the driver?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,38 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+        /// Configures the pattern start label for subsequent pattern execution.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.PatternControl.StartLabel = startLabel;
```
NikhilSivadas-NIB, line 124: Just to confirm, this follows the spec, but when siteNumbers is null, is the intended behavior to update only StartLabel without calling ConfigurePatternBurstSites()?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,38 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+
+                if (siteNumbers != null && siteNumbers.Length > 0)
+                {
+                    string siteList = string.Join(",", siteNumbers.Select(sn => $"site{sn}"));
+                    sessionInfo.Session.PatternControl.ConfigurePatternBurstSites(siteList);
```
NikhilSivadas-NIB: Again, this follows the feature spec, but I had a question. Do we need to filter siteNumbers per session before passing them to ConfigurePatternBurstSites()? As per the API documentation, ConfigurePatternBurstSites() throws a SelectorNameException if the siteList contains invalid sites. If a session does not have all the requested sites, which could lead to the above exception right? Hope its a valid scenario that may occur. If we are not filtering per session, should we handle this exception? If we are not filtering sites before passing and it has invalid sites, it will bubble up to the client code. What do you think?

NikhilSivadas-NIB (commented):
Can you check the comments ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: yes it is the expected driver format.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: yes. That is the expected behaviour. Not to call the function if the site list is empty or not provided by user etc.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +238,63 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
             Assert.Equal(5, failCountResults.ExtractSite(0).Count);
         }
+
+        [Theory]
+        [InlineData("TwoDevicesWorkForTwoSitesSeparately.pinmap", "TwoDevicesWorkForTwoSitesSeparately.digiproj")]
+        [InlineData("OneDeviceWorksForOnePinOnTwoSites.pinmap", "OneDeviceWorksForOnePinOnTwoSites.digiproj")]
+        public void SessionsInitialized_CommitSucceeds(string pinMap, string digitalProject)
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+
```
NikhilSivadas-NIB: Suggestion( Formatting) Since we are testing the Commit method here, var sessionsBundle = sessionManager.Digital("C0"); looks like part of the Arrange step. Could we add a blank line after the Arrange section to better follow AAA formatting?  also, can we add "_" after "Commit" in test name like other tests ?

NikhilSivadas-NIB: Similar suggestion for other tests, it would improve readability if we consistently separate Arrange, Act, and Assert, with only the method under test in the Act section.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +238,63 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+            Assert.All(patternStartLabel, label => Assert.Equal(expectedStartLabel, label));
+        }
+
+        [Theory]
+        [InlineData("TwoDevicesWorkForTwoSitesSeparately.pinmap", "TwoDevicesWorkForTwoSitesSeparately.digiproj")]
+        [InlineData("OneDeviceWorksForOnePinOnTwoSites.pinmap", "OneDeviceWorksForOnePinOnTwoSites.digiproj")]
+        public void SessionsInitialized_ConfigurePatternWithSiteNumbers_GetPatternStartLabelReturnsConfiguredLabel(string pinMap, string digitalProject)
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+
```
NikhilSivadas-NIB, line 305: For ConfigurePatternWithSiteNumbers, we are currently asserting only the StartLabel. Is there a way to also verify that ConfigurePattern is applied to the intended sites?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +238,63 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+        [InlineData("OneDeviceWorksForOnePinOnTwoSites.pinmap", "OneDeviceWorksForOnePinOnTwoSites.digiproj")]
+        public void SessionsInitialized_ConfigurePatternWithoutSpecifyingSites_GetPatternStartLabelReturnsConfiguredLabel(string pinMap, string digitalProject)
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+
+            var sessionsBundle = sessionManager.Digital(new string[] { "C0", "C1" });
+            const string expectedStartLabel = "TX_RF";
+            sessionsBundle.ConfigurePattern(expectedStartLabel);
+
+            var patternStartLabel = sessionsBundle.GetPatternStartLabel();
```
NikhilSivadas-NIB, line 277: Similar to the below comment, when siteNumber is not passed to ConfigurePattern, I believe the intended behavior is to apply the configuration to all sites. From the current implementation, it looks like only the start label is being set, and the underlying ConfigurePattern may not actually be invoked for all sites. If we are making changes there , should we also add a new test case to verify that ConfigurePattern is applied to all sites in the session? will that be possible to verify?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +238,63 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+        public void SessionsInitialized_CommitAndInitiateAndAbortPattern_Succeeds(string pinMap, string digitalProject)
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+
+            var sessionsBundle = sessionManager.Digital("C0");
+            const string expectedStartLabel = "TX_RF";
+            sessionsBundle.ConfigurePattern(expectedStartLabel);
+            sessionsBundle.Commit();
+            sessionsBundle.Initiate();
+            sessionsBundle.AbortPattern();
```
NikhilSivadas-NIB, line 263: Do we want to add an explicit no exception thrown assertion in this and next test case for better readability? or keep it implicit since the test already fails if Commit throws Exception

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +238,63 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+
+            var sessionsBundle = sessionManager.Digital(new string[] { "C0", "C1" });
+            const string expectedStartLabel = "TX_RF";
+            sessionsBundle.ConfigurePattern(expectedStartLabel, new int[] { 0, 1 });
+
+            var patternStartLabel = sessionsBundle.GetPatternStartLabel();
+
+            Assert.All(patternStartLabel, label => Assert.Equal(expectedStartLabel, label));
```
NikhilSivadas-NIB: Also, if we are addressing validation related changes in the implementation, could we consider adding negative test cases to cover invalid inputs or error scenarios?

NikhilSivadas-NIB (commented):
I have reviewed unit test as well, added some comments, can you have a look?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
NikhilSivadas-NIB: i think negative site number is valid for some system pins, in that case consider only check for duplicate site number if that add value in real scenarios.

NikhilSivadas-NIB (commented):
Just minor comments updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: Updated the spec

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Pattern is loaded if the `StartLabel` is valid. It is like a identifier for the loaded pattern. So that is the only way to verify the pattern is loaded.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: I looked for APIs to verify the sites being bursted. I did not find one so i think there is no way to verify programatically.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: Added test cases for empty and null cases.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: Added test case for duplicate site numbers

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: yes.I agree added the logic to filter the sites that are applicable to the current session.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Updated all the tests as per AAA

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Yes added validation logic and negative tests for the same.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Our unit tests will fail anyway if an exception is thrown - we have not put explicit Assertion for the same in the past.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.resx
```diff
@@ -168,4 +168,7 @@
   <data name="DCPower_GangedPinGroupDetected" xml:space="preserve">
     <value>This feature is not supported on a ganged pin group.</value>
   </data>
+  <data name="Digital_InvalidSites" xml:space="preserve">
+    <value>The specified sites are not present in the bundle.</value>
+  </data>
```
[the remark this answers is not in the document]
nitamil: Fixed it.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +103,45 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+            }
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.PatternControl.StartLabel = startLabel;
+                if (siteNumbers != null && siteNumbers.Length > 0)
+                {
+                    var filteredSites = sessionInfo.AssociatedSiteList.Where(sn => siteNumbers.Contains(sn)).ToArray();
+                    string siteList = string.Join(",", filteredSites.Select(sn => $"site{sn}"));
+                        sessionInfo.Session.PatternControl.ConfigurePatternBurstSites(siteList);
+                }
```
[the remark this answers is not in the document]
nitamil: Fixed it.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +103,45 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
+        {
+            var sitePinList = sessionsBundle.AggregateSitePinList;
+            var listOfSiteNumbers = sitePinList.Select(spi => spi.SiteNumber).Distinct().ToArray();
+            var invalidSites = siteNumbers?.Except(listOfSiteNumbers).ToArray();
+            if (invalidSites != null && invalidSites.Length > 0)
+            {
+                throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_InvalidSites, string.Join(", ", invalidSites)));
+            }
```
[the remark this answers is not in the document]
nitamil: Fixed it.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +103,48 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+
+        /// <summary>
+        /// Configures the pattern start label for subsequent pattern execution.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
+        {
+            var sitePinList = sessionsBundle.AggregateSitePinList;
```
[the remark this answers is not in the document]
nitamil: Fixed as per lower version of dotnet.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -113,6 +158,31 @@ public static PinSiteData<long> GetFailCount(this DigitalSessionsBundle sessions
...
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <returns>An array of the currently configured pattern start labels, one value per session.</returns>
+        public static string[] GetPatternStartLabel(this DigitalSessionsBundle sessionsBundle)
+        {
+            return sessionsBundle.DoAndReturnPerInstrumentPerChannelResults(sessionInfo =>
+            {
+                return sessionInfo.Session.PatternControl.StartLabel;
+            });
+        }
```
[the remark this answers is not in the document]
nitamil: Fixed

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
NikhilSivadas-NIB: @Mattjet27 @lornazh @nitamil 
According to the Digital driver API reference, when siteList is an empty string in ConfigurePatternBurstSites, all sites should be configured for pattern bursting. However, based on the feature spec and the current implementation, when siteNumbers is null or not provided, we only set the StartLabel and do not call ConfigurePatternBurstSites with an empty string.

This seems like a behavioral difference to me. Is this discrepancy intentional, or should we align the implementation with the API documentation?

Image

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
NikhilSivadas-NIB: Sorry, asking for my clarification, I see test cases added for empty start label, but empty startLabel valid case that we let user to set? My intention was to add a validation check in implementation for this case, but if its a valid case then its fine

NikhilSivadas-NIB (commented):
Added one comment to get clarification from Matt and Lorna,

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +100,49 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+        /// Configures the pattern start label for subsequent pattern execution.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
+        public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel, int[] siteNumbers = null)
+        {
+            startLabel = startLabel ?? string.Empty;
+            var sitePinList = sessionsBundle.AggregateSitePinList;
+            var listOfSiteNumbers = sitePinList.Select(spi => spi.SiteNumber).Distinct().ToArray();
```
NikhilSivadas-NIB: Can we validate siteNumbers per session inside Do and throw, NISemiconductorTestException there? Since driver calls are made per session, session-level validation seems more appropriate than aggregate bundle-level validation. If we fail fast on invalid session sites, we likely won’t need the additional filtering step.

NikhilSivadas-NIB (commented):
Added few comments

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: yes it valid case for the driver.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: This is bit tricky. This is related to comment in R133. Plese find the details there.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: Let say we have 1 pin and 2 sites that are mapped to 2 cards i.e site0/pinA is mapped to first PXI Digital card IO and the site1/pinA is mapped to 2nd PXI Digital card. Then there are 2 possible pinmaps based on the Group Value being set. 
Image

Case 1: Same Group (By Default: `Digital` group)
    Only one session will be created where both sites are in the session. In this case if we disable/enable one of the site by providing the site number it will work fine without any issues. 

Case 2: Different Group (say `Digital1` for first card and `Digital2` for the second card)
     Our abstraction layer will create 2 sessions as per the inputs from the TSM layer. 
one session contains `site0/PinA` and the other conains session contains `site1/PinA`. To FilterSites if user provided site number `1`. So we are selecting site number in the session where it makes applicable. 

Hope this helps.

Mattjet27: @NikhilSivadas-NIB, this was intentional. By design, we want a code path that does not actually invoke the underlying driver call, `ConfigurePatternBurstSites`, and would skip over that code, allowing any pre-existing configure sites to be used. We should still be allowing the user to pass in an empty string `""`, in which our extension method would reflect the same behavior as the underlying driver call, since the empty string will just get pass down directly to the `ConfigurePatternBurstSites` method. Now, since we are using an `int[]` for this parameter in the extension method, I suppose we'd have to allow an empty array to be passed in for the equivalent behavior. The current code implementation does not allow for this. We can also consider revising this design so that it does in fact always make the driver call and pass it in an empty string: `ConfigurePatternBurstSites("")`.

That said, I think to @nitamil's point, it would be possible (although NEVER advisable, as it goes against pin mapping best practices for NI Digital) for the sites to be split across different device sessions for the pins used in a pattern. In such a scenario configuring the pattern burst for all sites by passing the `ConfigurePatternBurstSites` driver method an empty string, I do not believe would interfere with the other session but passing it a `siteX/PinY` pair for a site that is not mapped to the instrument session would likely result in an error.

I am imaging that users will pass the  `ISemiconductorContext.SiteNumbers.SiteNumbers` collection directly to the `siteNumbers` argument when invoking `ConfigurePattern` extension method AND wanting to burst across all sites. Therefore, it would be nice if instead of `int[]` the `siteNumbers` parameter could be and `ICollection<int>` so that it can either be pass in as a `int[]` or directly accept the `ISemiconductorContext.SiteNumbers.SiteNumbers` collection. It must be up to the implementation to ensure the appropriate sites are set based on which sites are actually mapped to each of the sessions in the bundle when in the above scenario.

nitamil: Updated to ICollection<int> and allowed the Empty string to be set when empty collection is passed.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +101,46 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+                throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_InvalidSites, string.Join(", ", invalidSites)));
+            }
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.PatternControl.StartLabel = startLabel;
+                if (siteNumbers != null)
+                {
+                    var filteredSites = sessionInfo.AssociatedSiteList.Where(sn => siteNumbers.Contains(sn)).ToArray();
+                    var siteList = string.Join(",", filteredSites.Select(sn => $"site{sn}"));
+                    sessionInfo.Session.PatternControl.ConfigurePatternBurstSites(siteList);
```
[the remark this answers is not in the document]
nitamil: This is by design. Not going to fix it.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
NikhilSivadas-NIB: @Mattjet27 Thanks for the clarification. I understand the intent now.

Since site filtering is already performed per session before the driver call, I don't think additional validation that throws NISemiconductorTestException is required at bundle level at the beginning of method . 

With the discussion,  behavior now defined as follows:

- siteNumbers == null : do not call ConfigurePatternBurstSites() and update only StartLabel
- Empty collection : call ConfigurePatternBurstSites("") to explicitly configure all sites.
- Non empty collection : configure only the filtered, session applicable sites.

Given this, @nitamil  can we remove the current site validation that throws NISemiconductorTestException and rely on the per session filtering logic instead ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
```diff
@@ -100,6 +101,46 @@ public static SiteData<bool> GetSitePassFail(this DigitalSessionsBundle sessions
...
+                sessionInfo.Session.PatternControl.Commit();
+            });
+        }
+
+        /// <summary>
+        /// Configures the pattern start label for subsequent pattern execution.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="startLabel">The pattern name or exported pattern label to configure.</param>
+        /// <param name="siteNumbers">The site numbers to configure the pattern for.</param>
```
NikhilSivadas-NIB: do we need to update the behavior in doc?

NikhilSivadas-NIB (commented):
Minor comments added, can you check?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: Updated the documentation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs
nitamil: As discussed added a null check for the parameter.

lornazh: @Mattjet27, @NikhilSivadas-NIB and I had a discussion yesterday regarding this extension method.

Can we step back a bit, because I'd like to understand why do we need to take a `siteNumbers` parameter for this extension method. Does it conflict with the associated site list information inside the bundle object? If the user want to only configure a subset site list, would filtering the bundle object by site and then call this extension method conform better to how STL's extension methods are normally used? I think we normally would like to avoid passing pin or site information directly from extension methods (I don't think there is an exception in current code base yet), because bundle object should be the entity that manages pin/site that each method operates.

Even though we determined to go this path, @nitamil I don't think the current invalid site validation logic, which is around `AggregateSitePinList`, is correct. I think it's possible that different `sessionInfo` may associate with different set of sites, so the check should happen session by session using the `AssociatedSiteList` property.

One other concern is, we should be very careful using the empty string as site list or channel string, because empty string means all sites/channels associated with the session to driver, but the bundle object is very likely to contain only a subset sites/channels. So the empty string customers pass in should be interpreted as `AssociatedSiteList`, `AssociatedSitePinList`, `AllChannelsString`, `AllChannelsOutput` instead (I believe example logic could be found in our existing DCPower extension method implementation).

nitamil: Hi @Mattjet27 , I marked this thread as unresolved. Based on lorna's Input, my understanding is that it breaks the Design logic(abstraction) of STL. your input is needed to proceed further.

Mattjet27: @nitamil, @lornazh, @NikhilSivadas-NIB, as we discussed during today's call. I am aligned with Lorna's suggestion and reasoning. We can make it a requirement that the user must filter the bundle if they need to target specific sites. Therefore, let us proceed as follows:

1. Remove `siteNumbers` as an input parameter to the `ConfigurePattern` method.

   ```cs
   public static void ConfigurePattern(this DigitalSessionsBundle sessionsBundle, string startLabel)
   ```

2. Update the implementation for the `ConfigurePattern` method such that it always calls both `sessionInfo.Session.PatternControl.StartLabel = startLabel;` and `sessionInfo.Session.PatternControl.ConfigurePatternBurstSites(siteList);`, where the `siteList` passed into `ConfigurePatternBurstSites` is obtained from the bundle itself.

3. We must test the scenarios where the user calls this on a site-filtered bundle and then calls the Initiate method either A. on the same site-filtered bundle, or B. on the original bundle object, and confirm the expected behavior: does this result in an Exception being through by the driver? does the pattern burst on different sites depending on which bundle the Initiate operation is performed on? Note that this testing may warrant revisiting this decision.

nitamil: 1 and 2, I changed locally and for point 3 when i validated it is working as expected. i.e. 

Driver is throwing error when I query for `GetFailCount();` on the entire bundle. when i use the filtered bundle it is giving data. 

```csharp
            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
            var sessionsBundle = sessionManager.Digital(new string[] { "C0", "C1" });
            const string expectedStartLabel = "TX_RF";
            var bundleWithSite0 = sessionsBundle.FilterBySite(0);

            bundleWithSite0.ConfigurePattern(expectedStartLabel);
            bundleWithSite0.BurstPattern(expectedStartLabel);

            // var results = sessionsBundle.GetFailCount();
            var results = bundleWithSite0.GetFailCount();
            Assert.Single(results.SiteNumbers);
            Assert.Equal(2, results.ExtractSite(0).Count);
            Assert.Contains("C0", results.ExtractSite(0).Keys);
            Assert.Contains("C1", results.ExtractSite(0).Keys);
            // Assert.Equal(2, results.ExtractSite(1).Count);
            // Assert.Contains("C0", results.ExtractSite(1).Keys);
            // Assert.Contains("C1", results.ExtractSite(1).Keys);

Comment (nitamil) [reply SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs:124]: @Mattjet27,  Let me know if can proceed with 1,2 and add validation cases as per exceptions(negative tests).

Comment (Mattjet27) [reply SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs:124]: @nitamil, proceed.

Comment (nitamil) [reply SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/Pattern.cs:124]: Updated the test cases

File: SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs [hunk 19]
```diff
@@ -238,5 +236,135 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+        public void SessionsInitialized_ConfigurePatternAndInitiateOnFilteredSite_GetFailCountOnAllSitesThrowsException(string pinMap, string digitalProject)
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager(pinMap, digitalProject);
+            var sessionsBundle = sessionManager.Digital(new string[] { "C0", "C1" });
+            const string expectedStartLabel = "TX_RF";
+
+            var bundleWithSite0 = sessionsBundle.FilterBySite(0);
+            bundleWithSite0.ConfigurePattern(expectedStartLabel);
+            bundleWithSite0.Initiate();
+            void GetFailCountOnAllSites() => sessionsBundle.GetFailCount();
```
NikhilSivadas-NIB: can we move this line `void GetFailCountOnAllSites() => sessionsBundle.GetFailCount();` to last section as test name suggests?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
```diff
@@ -238,5 +236,135 @@ public void SessionsInitialized_BurstPatternWithoutSpecifyingPins_Succeeds()
...
+            Assert.Contains("C1", results.ExtractSite(0).Keys);
+            Assert.Equal(2, results.ExtractSite(1).Count);
+            Assert.Contains("C0", results.ExtractSite(1).Keys);
+            Assert.Contains("C1", results.ExtractSite(1).Keys);
+        }
+
+        [Theory]
+        [InlineData("TwoDevicesWorkForTwoSitesSeparately.pinmap", "TwoDevicesWorkForTwoSitesSeparately.digiproj")]
+        [InlineData("OneDeviceWorksForOnePinOnTwoSites.pinmap", "OneDeviceWorksForOnePinOnTwoSites.digiproj")]
+        public void SessionsInitializedSiteFiltered_ConfigurePatternAndInitiateOnFilteredSite_Succeeds(string pinMap, string digitalProject)
```
NikhilSivadas-NIB: As Copilot suggested can we rename the test ? currently both arrange and act part indicate sitefiltered session,  also move `var bundleWithSite0 = sessionsBundle.FilterBySite(0);` to mid section (Act part) of testcase so that naming is consistent with other test similar cases

NikhilSivadas-NIB (approved):
I have added two minor comments and approved the PR

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Updated.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/PatternTests.cs
nitamil: Updated.

PR #518, opened 2026-06-15
[POC]Adding FGen Extensions Design Prototype

hcnagaraj-e:
### Description

It is a POC PR, created to illustrate the class structure for the extension methods being added for FGen

### DIFF
--- a/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
+++ b/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
@@ -38,6 +38,7 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGe
                 sessionInfo.ConfigureChannels(associatedChannels);
 
                 sessionInfo.Session.InitiateGeneration();
+                var serialNumber = sessionInfo.Session.Identity.SerialNumber;
             });
         }
 
--- a/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
+++ b/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
@@ -9,30 +9,6 @@ namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGe
     public static class Control
     {
         /// <summary>
-        /// Configure output mode.
-        /// </summary>
-        /// <param name="sessionsBundle">The FGen sessionsBundle.</param>
-        /// <param name="outputMode">The output mode to configure.</param>
-        public static void ConfigureOutputMode(this FgenSessionsBundle sessionsBundle, OutputMode outputMode)
-        { }
-
-        /// <summary>
-        /// Configure output mode.
-        /// </summary>
-        /// <param name="sessionsBundle">The FGen sessionsBundle.</param>
-        /// <param name="outputMode">The output mode to configure.</param>
-        public static void ConfigureOutputMode(this FgenSessionsBundle sessionsBundle, SiteData<OutputMode> outputMode)
-        { }
-
-        /// <summary>
-        /// Configure output mode.
-        /// </summary>
-        /// <param name="sessionsBundle">The FGen sessionsBundle.</param>
-        /// <param name="outputMode">The output mode to configure.</param>
-        public static void ConfigureOutputMode(this FgenSessionsBundle sessionsBundle, PinSiteData<OutputMode> outputMode)
-        { }
-
-        /// <summary>
         /// Commit.
         /// </summary>
         /// <param name="sessionsBundle">The FGen sessionsBundle.</param>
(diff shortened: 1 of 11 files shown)


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Control class for controlling waveform generation operations.
+    /// </summary>
+    public static class Control
+    {
+        /// <summary>
+        /// Configure output mode.
+        /// </summary>
+        public static void ConfigureOutputMode()
```
Ashutosh-NI: As we are configuring the output mode, will it be better if we have it in `Output.cs`?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/TriggerAndEvents.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGen
+{
+    internal class TriggerAndEvents
```
Ashutosh-NI: Any specific reason for this class to be `internal`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Type.cs
```diff
@@ -0,0 +1,67 @@
...
+        RamDown,
+        DC,
+        Noise,
+        User
+    }
+
+    /// <summary>
+    /// Type class contains all type definitions needed for FGen operations.
+    /// </summary>
+    public class StdWaveformSettings
```
Ashutosh-NI: I think we should name the class as `StandardWaveformSettings` which will be much more consistent with the other names.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/SampleExample.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            /*
+             * ConfigureOutputMode
+             * ConfigureStandardWaveform
+             * Initiate
+             * Poll for IsDone
+             * Abort
+             */
+        }
+    }
```
Ashutosh-NI: This will be part of the examples right ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/StdWaveform.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public static class StdWaveform
+    {
+        /// <summary>
+        /// Confgure standard waveform parameters.
+        /// </summary>
+        public static void ConfigureStdWaveform()
+        { }
+    }
+}
```
Ashutosh-NI: Can we replace the `Std` with `Standard` in naming as that might be consistent with the other names.

Ashutosh-NI (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Control.cs
hcnagaraj-e: It is part of operation control API, user will select standard wave form generation option.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Type.cs
hcnagaraj-e: @Ashutosh-NI  I referred DAQmx Extension layer implementation. They do have custom class objects created for handling waveform parameters. We will end up adding more settings classes under this file in part2. 
This is a different Type.cs available in Extensions.

DAQmx example: https://github.com/ni/semi-test-library-dotnet/blob/main/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DAQmx/Types.cs

hcnagaraj-e: Please feel free to re-open comment if it still does not make sense.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/SampleExample.cs
hcnagaraj-e: Yes, it is an example code to illustrate how extensions will be called. This code will not be added in implementation, instead actual code snippet and sequence style examples are added in example directory

I can add some comments in the top to clarify this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/StdWaveform.cs
hcnagaraj-e: Noted: I have used StdWaveform in all places. I will fix it if there are discrepancies

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/TriggerAndEvents.cs
hcnagaraj-e: @Ashutosh-NI I was trying to avoid errors thrown for not having documentation for public classes. 
I have changed it to public and added summary for class and its methods.

PR #526, opened 2026-07-01
Add TMU configXX methods

Ashutosh-NI:
### Why should this be merged?

The STL Digital TMU surface API was missing configure methods for five common signal measurement types rise time, fall time, duty cycle, pulse width, and skew. Without these, users have to manually set individual TMU attributes (start source, stop source, event, polarity) to perform these measurements. This PR closes that gap by providing a single method per measurement type that correctly sets all required attributes.

### What changed?

**`TmuExtensions.cs`**
- Added `ConfigureTMURiseTimeMeasurement`
- Added `ConfigureTMUFallTimeMeasurement`
- Added `ConfigureTMUDutyCycleMeasurement` - accepts `TmuDutyCycle.High` or `TmuDutyCycle.Low`
- Added `ConfigureTMUPulseWidthMeasurement` - accepts `TmuPulseWidth.High` or `TmuPulseWidth.Low`
- Added `ConfigureSkewMeasurement(string, string, ...)` - single reference/target pin pair
- Added `ConfigureSkewMeasurement(string[], string[], ...)` - multiple paired reference/target pins

### What testing was done?
Added unit test for the all the mentioned extension methods.
All tests run on tester hardware.
image


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,302 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        /// <param name="targetPinNames">The pins to use as the target (stop) source for the skew measurement.</param>
+        /// <param name="edgeType">The type of edge to detect. Only accepts <see cref="TmuPolarity.RisingEdge"/> or <see cref="TmuPolarity.FallingEdge"/>.</param>
+        /// <param name="samplesToAcquire">The number of samples to acquire for the TMU measurement.</param>
+        /// <param name="armType">
+        /// The type of signal used to arm the TMU measurement.<br/>
+        /// The TMU's arm input is used to frame, or select, the start and stop events of interest for each TMU sample.
+        /// </param>
+        public static void ConfigureSkewMeasurement(
+            this DigitalSessionsBundle sessionsBundle,
+            string[] referencePinNames,
```
NikhilSivadas-NIB, line 367: Why this signature different from feature spec?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,302 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+            string[] targetPinNames,
+            TmuPolarity edgeType,
+            long samplesToAcquire,
+            TmuArmType armType = TmuArmType.Immediate)
+        {
+            // Validate array lengths match
+            if (referencePinNames.Length != targetPinNames.Length)
```
NikhilSivadas-NIB: referencePinNames.Length and targetPinNames.Length could throw NullRef Exception now if user pass null, can we add a check ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,302 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+                        .FirstOrDefault(sp => sp.PinName == targetPinName && sp.SiteNumber == sitePinInfo.SiteNumber);
+                    if (targetSitePinInfo == null)
+                    {
+                        throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUSkewTargetPinNotFound, targetPinName, sitePinInfo.SiteNumber));
+                    }
+
+                    ConfigureSkewMeasurementForSitePin(
+                        sessionInfo,
+                        sitePinInfo,
+                        targetSitePinInfo,
```
NikhilSivadas-NIB: targetSitePinInfo parameter is accepted as targetSitePinInfo in calling method, can we reorder?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,302 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        /// <param name="samplesToAcquire">The number of samples to acquire for the TMU measurement.</param>
+        /// <param name="armType">
+        /// The type of signal used to arm the TMU measurement.<br/>
+        /// The TMU's arm input is used to frame, or select, the start and stop events of interest for each TMU sample.
+        /// </param>
+        public static void ConfigureSkewMeasurement(
+            this DigitalSessionsBundle sessionsBundle,
+            string[] referencePinNames,
+            string[] targetPinNames,
+            TmuPolarity edgeType,
```
NikhilSivadas-NIB, line 369: Can we validate edgeType at the API entry point to provide earlier detection of invalid input scenarios?

NikhilSivadas-NIB (commented):
Reviewed partially, added few comments please have a look

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: We are already handling the `edgeType` check within the `ConfigureSkewMeasurementForSitePin` helper function.

If we move this check to `ConfigureSkewMeasurement`, it could introduce some code duplication. Specifically, `ConfigureSkewMeasurementForSitePin` would still need to determine and assign the appropriate `TmuSourceEvent` based on the `edgeType`, which would likely require another switch (or equivalent conditional logic).

Keeping the `edgeType` handling within `ConfigureSkewMeasurementForSitePin` helps keep the related logic consolidated in one place and avoids repeating the same decision-making logic across methods.

@NikhilSivadas-NIB, please let me know your thoughts on this approach.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Added a helper function to validate all the parameters and throw appropriate exception.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: Got your point, then can we move this code to a helper method? because it looks currently the following code is already duplicated in both **ConfigureSkewMeasurementForSitePin** and **ConfigurePeriodMeasurementForSitePin**, if we move to helper we could call it from all the places, so we could avoid code dupication and also early detect  invalid input at the entry point of public APIs ?
`TmuSourceEvent sourceEvent;
switch (edgeType)
{
    case TmuPolarity.RisingEdge:
        sourceEvent = TmuSourceEvent.Voh;
        break;
    case TmuPolarity.FallingEdge:
        sourceEvent = TmuSourceEvent.Vol;
        break;
    default:
        throw new ArgumentOutOfRangeException(nameof(edgeType), edgeType, string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUUnsupportedPolarity));
}`

NikhilSivadas-NIB (commented):
Replied to comment, please check

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated the parameters .

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Created a helper function and used it in both period measurement and skew measurment.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: I will be creating the PR to update the same in the spec.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,291 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        public static void ConfigureTMUSkewMeasurement(
+            this DigitalSessionsBundle sessionsBundle,
+            string[] referencePinNames,
+            string[] targetPinNames,
+            TmuPolarity edgeType,
+            long samplesToAcquire,
+            TmuArmType armType = TmuArmType.Immediate)
+        {
+            ValidateSkewParameters(referencePinNames, targetPinNames, armType, sessionsBundle.Pins);
+
```
NikhilSivadas-NIB, line 375: Can we also call the newly added helper method GetSourceEventForEdge here as well to throw early when edge type is invalid? Also consider renaming the helper method to "ValidateAndGetSourceEventForEdge".

Otherwise without avoid rename of the helper,  we could call " var sourceEvent = GetSourceEventForEdge(edgeType);" after"ValidateSkewParameters" and pass sourceEvent to ConfigureTMUSkewMeasurementForSitePin, the only concern in that case is its already having mnay parameters. Let me know your thoughts

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -968,6 +944,19 @@ private static void ConfigureSkewMeasurementForSitePin(
...
+        private static TmuSourceEvent GetSourceEventForEdge(TmuPolarity edgeType)
+        {
+            switch (edgeType)
+            {
+                case TmuPolarity.RisingEdge:
+                    return TmuSourceEvent.Voh;
+                case TmuPolarity.FallingEdge:
+                    return TmuSourceEvent.Vol;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(edgeType), edgeType, string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUUnsupportedPolarity));
```
NikhilSivadas-NIB: we usually wrap it with STL specific exception right?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +221,291 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+            {
+                referenceToTargetMap[referencePinNames[i]] = targetPinNames[i];
+            }
+
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            {
+                // Configure only for reference pins (which have the TMU assigned)
+                if (referenceToTargetMap.TryGetValue(sitePinInfo.PinName, out string targetPinName))
+                {
+                    // Find the target pin's channel string for the same site
```
NikhilSivadas-NIB: are we actually finding target pins channel string here? we are getting sitePinInfo of target pin and site right?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +229,291 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        /// - <see cref="TmuAttributes.TmuStartSourceEventPolarity"/> = <see cref="TmuPolarity.RisingEdge"/><br/>
+        /// - <see cref="TmuAttributes.TmuStopSource"/> = same pin as start source<br/>
+        /// - <see cref="TmuAttributes.TmuStopSourceEvent"/> = <see cref="TmuSourceEvent.Voh"/><br/>
+        /// - <see cref="TmuAttributes.TmuStopSourceEventPolarity"/> = <see cref="TmuPolarity.RisingEdge"/><br/>
+        /// - <see cref="TmuAttributes.TmuSamplesToAcquire"/> = value of <paramref name="samplesToAcquire"/> parameter.<br/>
+        /// - <see cref="TmuAttributes.TmuArmType"/> = value of <paramref name="armType"/> parameter.
+        /// </para>
+        /// <para>
+        /// This measurement requires 2 comparators.<br/>
+        /// This method does not enable the TMU (see <see cref="TmuAttributes.TmuEnabled"/>).
```
NikhilSivadas-NIB: inline doc says, it does not enable TMU, but in private method we are enabling, is it intentional?

NikhilSivadas-NIB: same for other configure methods

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -642,6 +924,152 @@ private static void ConfigurePeriodMeasurementForSitePin(DigitalSessionInformati
...
 
+        private static void ConfigureTMUSkewMeasurementForSitePin(
+            DigitalSessionInformation sessionInfo,
+            SitePinInfo referenceSitePinInfo,
+            SitePinInfo targetSitePinInfo,
+            TmuPolarity edgeType,
+            long samplesToAcquire,
+            TmuArmType armType)
+        {
+            TmuSourceEvent sourceEvent = GetSourceEventForEdge(edgeType);
```
NikhilSivadas-NIB: Sorry, I see the following code in the new private methods kind of duplicate across all configure methods? can we also move this also to a helper if we could pass the parameter based on each parameter based on overload?

NikhilSivadas-NIB (commented):
I have added some reviews, can you check?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: can we have a helper like this? so most of the public APIs could call this directly without separate private helper method for each public API, thoughts ? ConfigureDutyCycleMeasurementForSitePin and ConfigurePulseWidthMeasurementForSitePin could still have private method with switch based on those enums

 ```
private static void ConfigureTmuInternal(
     DigitalSessionInformation sessionInfo,
     SitePinInfo sitePinInfo,
     TmuSourceEvent startEvent,
     TmuPolarity startPolarity,
     TmuSourceEvent stopEvent,
     TmuPolarity stopPolarity,
     long samplesToAcquire,
     TmuArmType armType)
 {
     var tmuContext = (sitePinInfo as DigitalSitePinInfo).AssignedTmuContext;
     var tmu = GetDigitalTmus(sessionInfo.Session).GetTmu(tmuContext);

     tmu.Start.Source = sitePinInfo.IndividualChannelString;
     tmu.Start.SourceEvent = startEvent;
     tmu.Start.SourceEventPolarity = startPolarity;

     tmu.Stop.Source = sitePinInfo.IndividualChannelString;
     tmu.Stop.SourceEvent = stopEvent;
     tmu.Stop.SourceEventPolarity = stopPolarity;

     tmu.SamplesToAcquire = samplesToAcquire;
     tmu.ArmType = armType;
     tmu.Enabled = true;
 }
```

File: SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs [hunk 10]
```diff
@@ -221,6 +229,291 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        /// </remarks>
+        /// <param name="sessionsBundle">The <see cref="DigitalSessionsBundle"/> object.</param>
+        /// <param name="referencePinNames">The pins to use as the reference (start) source for the skew measurement.</param>
+        /// <param name="targetPinNames">The pins to use as the target (stop) source for the skew measurement.</param>
+        /// <param name="edgeType">The type of edge to detect. Only accepts <see cref="TmuPolarity.RisingEdge"/> or <see cref="TmuPolarity.FallingEdge"/>.</param>
+        /// <param name="samplesToAcquire">The number of samples to acquire for the TMU measurement.</param>
+        /// <param name="armType">
+        /// The type of signal used to arm the TMU measurement.<br/>
+        /// The TMU's arm input is used to frame, or select, the start and stop events of interest for each TMU sample.
+        /// </param>
```
NikhilSivadas-NIB, line 357: this could throw exception on validation right? can we update inline doc

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -221,6 +229,291 @@ public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle session
...
+        /// </para>
+        /// <para>
+        /// - <see cref="TmuAttributes.TmuSamplesToAcquire"/> = value of <paramref name="samplesToAcquire"/> parameter.<br/>
+        /// - <see cref="TmuAttributes.TmuArmType"/> = value of <paramref name="armType"/> parameter.
+        /// </para>
+        /// <para>
+        /// This measurement requires 1 comparator.<br/>
+        /// This method does not enable the TMU (see <see cref="TmuAttributes.TmuEnabled"/>).
+        /// Call <see cref="EnableTMU"/> separately when ready to reserve the TMU resource.<br/>
+        /// The returned measurement value is the time duration, not a percentage.
```
NikhilSivadas-NIB: this line needs to be rephrased? this method does not return anything right?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -246,6 +247,201 @@ public void Inititalize_ConfigureTMUSampleTimeoutSucceeds(double timeout, bool u
...
+        public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsAndEitherEdge_ThrowsNISemiconductorTestException()
+        {
+            var sessionsBundle = InititalzeAndCreateBundle();
+
+            Assert.Throws<NISemiconductorTestException>(() =>
+                sessionsBundle.ConfigureTMUSkewMeasurement(
+                    new[] { "C0" }, new[] { "C1" }, TmuPolarity.EitherEdge, 1));
+            sessionsBundle.DisableTMU();
             sessionsBundle.ClearTMUAssignment();
         }
```
NikhilSivadas-NIB, line 681: can we add some test cases for "ConfigureTMUSkewMeasurement" to add coverage to negative validation cases? For example 
1. reference/target pin counts differs?
2. reference/target PinNames = null
3. unsupported edgeType 
4. pinNames not in bundle etc ..

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -246,6 +247,201 @@ public void Inititalize_ConfigureTMUSampleTimeoutSucceeds(double timeout, bool u
...
+        {
+            var sessionsBundle = InititalzeAndCreateBundle();
+
+            sessionsBundle.ConfigureTMUSkewMeasurement(new[] { "C0" }, new[] { "C1" }, edgeType, 1);
+            sessionsBundle.DisableTMU();
+            sessionsBundle.ClearTMUAssignment();
+        }
+
+        [Fact]
+        public void Inititalize_ConfigureSkewMeasurementWithOverlappingReferencAndTargetPinsThrowsNISemiconductorTestException()
```
NikhilSivadas-NIB: Looks underscore missed before assert part in test name

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -246,6 +247,201 @@ public void Inititalize_ConfigureTMUSampleTimeoutSucceeds(double timeout, bool u
...
+            sessionsBundle.DisableTMU();
+            sessionsBundle.ClearTMUAssignment();
+        }
+
+        [Fact]
+        public void Inititalize_ConfigureTMURiseTimeMeasurementWithArmTypeSucceeds()
+        {
+            var sessionsBundle = InititalzeAndCreateBundle();
+
+            sessionsBundle.ConfigureTMURiseTimeMeasurement(samplesToAcquire: 1);
```
NikhilSivadas-NIB, line 326: 1. can we add tests for negative cases for "ConfigureTMURiseTimeMeasurement" like passing pins that's not available in bundle, pins=null (configure all pins in bundle)

Consider the same for other similar newly added methods

NikhilSivadas-NIB (commented):
Added some more comments, please have a look

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: This will get wrapped in `NISemiconductorTestException` as when `Do` will catch this exception it will take it and wrap it in `NISemiconductorTestException`. So end user will see this as exception with inner exception as `ArgumentOutOfRangeException`.

I can change it to throw `NISemiconductorTestException` but i thought `ArgumentOutOfRangeException` might give more information related to argument which is failing.
@NikhilSivadas-NIB  let me know your thought on this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Renamed the method to `ValidateAndGetSourceEventForEdge` and added it just after `ValidateSkewParameters` for early validation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Added `NISemiconductorTestException` as now the check is happenign outside of `Do`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: opps👀!! missed it , updated it according,

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Removed this from each of the configure methods.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Added this common helper for all the methods and used it accordingly.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Removed

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated the inline doc to mention the error/exception thrown by the methods.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Added the tests for the negative validations

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Updated the name.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Added test for negative scenarios for all the methods.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -209,14 +218,395 @@ public static void ClearTMUAssignment(this DigitalSessionsBundle sessionsBundle,
...
+                                startEvent: TmuSourceEvent.Vol,
+                                startPolarity: TmuPolarity.FallingEdge,
+                                stopSource: channel,
+                                stopEvent: TmuSourceEvent.Vol,
+                                stopPolarity: TmuPolarity.RisingEdge,
+                                samplesToAcquire: samplesToAcquire,
+                                armType: armType);
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(pulseWidthType), pulseWidthType, string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUUnsupportedPolarity));
```
NikhilSivadas-NIB: Inline doc aligns with NISemiconductorTestException, can we update?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -246,6 +247,436 @@ public void Inititalize_ConfigureTMUSampleTimeoutSucceeds(double timeout, bool u
...
+        }
+
+        #endregion
+
+        #region Configure Skew Measurement (Multiple Pin Pairs) Tests
+
+        [Theory]
+        [InlineData(TmuPolarity.RisingEdge)]
+        [InlineData(TmuPolarity.FallingEdge)]
+        public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsSucceeds(TmuPolarity edgeType)
```
NikhilSivadas-NIB: MultiplePins sounds confusing, we input only single pin for both ref and target here right. Instead can we call Unique or District pins? Same with next to next test case

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -209,14 +218,395 @@ public static void ClearTMUAssignment(this DigitalSessionsBundle sessionsBundle,
         /// The TMU's arm input is used to frame, or select, the start and stop events of interest for each TMU sample.
         /// </param>
         /// <param name="pinNames">The specific pins to configure the TMU for. When <c>null</c>, all pins are targeted.</param>
+        /// <exception cref="NISemiconductorTestException">
+        /// Thrown when one or more of the requested <paramref name="pinNames"/> are not present in the sessions bundle,
+        /// or when the <paramref name="edgeType"/> is <see cref="TmuPolarity.EitherEdge"/> (an unsupported polarity).
+        /// </exception>
         public static void ConfigurePeriodMeasurement(this DigitalSessionsBundle sessionsBundle, TmuPolarity edgeType, long samplesToAcquire, TmuArmType armType = TmuArmType.Immediate, string[] pinNames = null)
```
NikhilSivadas-NIB, line 274: Skew measurement APIs has Arm Type validation , shall we add the same in the other newly added APIs to keep consistency ?

NikhilSivadas-NIB, line 274: typo in doc, 
`/// - <see cref="TmuAttributes.TmuStopSource"/> = same pin a start source<br/>`

NikhilSivadas-NIB, line 274: inline do says"<see cref="TmuAttributes.TmuEnabled"/> = <c>true</c>"

However not there in other public methods

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -209,14 +218,395 @@ public static void ClearTMUAssignment(this DigitalSessionsBundle sessionsBundle,
...
+        /// <exception cref="NISemiconductorTestException">Thrown when one or more of the requested <paramref name="pinNames"/> are not present in the sessions bundle.</exception>
+        public static void ConfigureTMUFallTimeMeasurement(this DigitalSessionsBundle sessionsBundle, long samplesToAcquire, TmuArmType armType = TmuArmType.Immediate, string[] pinNames = null)
+        {
+            ValidatePinsOfTMU(sessionsBundle.Pins, pinNames);
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            {
+                if (DoForThisPin(pinNames, sitePinInfo.PinName))
+                {
+                    string tmuContext = (sitePinInfo as DigitalSitePinInfo).AssignedTmuContext;
+                    DigitalTmu tmu = GetDigitalTmus(sessionInfo.Session).GetTmu(tmuContext);
```
NikhilSivadas-NIB: GetAssignedTmu method could be used here right?

NikhilSivadas-NIB (commented):
Can you check it overall again? 
 I am not sure if there was merged from any other story branch , I am unable to view any changes from my last review, it says no commits after that, please have a look

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Replaced  `ArgumentOutOfRangeException` with `NISemiconductorTestException` in the implementation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Added validation for `ArmType` at all the required places.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated the this and other required places to use `GetAssignedTmu`

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: added for other method

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Fixed the typeo.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -50,6 +50,14 @@ public static void TMUInitiate(this DigitalSessionsBundle sessionsBundle, string
             });
         }
 
+        /// <inheritdoc cref="TMUInitiate(DigitalSessionsBundle, string[])"/>
+        /// <param name="sessionsBundle"/>
+        /// <param name="pinName">The specific pin to initiate the TMU measurement on.</param>
+        public static void TMUInitiate(this DigitalSessionsBundle sessionsBundle, string pinName)
```
NikhilSivadas-NIB, line 56: shall we validate pinName here as well? and also in other APIs which take single pinNAme as input ? We are validating pinNames in its overlaod which takes array of pinName

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -73,6 +81,14 @@ public static void EnableTMU(this DigitalSessionsBundle sessionsBundle, string[]
             });
         }
 
+        /// <inheritdoc cref="EnableTMU(DigitalSessionsBundle, string[])"/>
+        /// <param name="sessionsBundle"/>
+        /// <param name="pinName">The specific pin for which to enable the assigned TMU resource.</param>
+        public static void EnableTMU(this DigitalSessionsBundle sessionsBundle, string pinName)
```
NikhilSivadas-NIB, line 87: here as well

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -209,18 +266,484 @@ public static void ClearTMUAssignment(this DigitalSessionsBundle sessionsBundle,
...
+                                startEvent: TmuSourceEvent.Vol,
+                                startPolarity: TmuPolarity.FallingEdge,
+                                stopSource: channel,
+                                stopEvent: TmuSourceEvent.Vol,
+                                stopPolarity: TmuPolarity.RisingEdge,
+                                samplesToAcquire: samplesToAcquire,
+                                armType: armType);
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(dutyCycleType), dutyCycleType, string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUUnsupportedPolarity));
```
NikhilSivadas-NIB: NISemiconductorTestException ?

NikhilSivadas-NIB: Looks like ResourceStrings.Digital_TMUUnsupportedPolarity has been used in multiple places for default case throw for different enums, this might confuse right?

NikhilSivadas-NIB (commented):
added few more comments, can you check ?

NikhilSivadas-NIB (commented):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: I have updated the existing `ValidateTMUPins` method to have the check for null or empty pin, as having separate validation just these methods might have  code repetition as all of these methods will go through `ValidateTMUPins`

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: I have updated the existing ValidateTMUPins method to have the check for null or empty pin, as having separate validation just these methods might have code repetition as all of these methods will go through ValidateTMUPins

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: updated to have `NISemiconductorTestException`

Ashutosh-NI: Updated the code to have sperate resource strings.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: Yes, i did not notice that, thats fine in that case

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: Sorry, just one more thing, i think we missed to add inline doc for that validation error, for both overload that take pinNames array and pinName.

Ashutosh-NI: @NikhilSivadas-NIB  Just to clarify, you mean the inline doc for the exception thrown for the validation of pins?

NikhilSivadas-NIB: Yes, that's right

Ashutosh-NI: Updated the code to have the exception information in all the methods.

NikhilSivadas-NIB: i think its still missing for single pin overloads

Ashutosh-NI: As single pin overloads are inheriting the doc form the pin as array overload, so the exception will also get inherited.

NikhilSivadas-NIB: but if we inherit, pinNames also gets inherited right? If yes, pinNames is not a parameter in single pin overload right? also message is not generic pinNames overload

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -681,5 +1393,59 @@ private static DigitalTmuCollections GetDigitalTmus(NIDigital session)
...
+        private static TmuSourceEvent ValidateAndGetSourceEventForEdge(TmuPolarity edgeType)
+        {
+            switch (edgeType)
+            {
+                case TmuPolarity.RisingEdge:
+                    return TmuSourceEvent.Voh;
+                case TmuPolarity.FallingEdge:
+                    return TmuSourceEvent.Vol;
+                default:
+                    throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUUnsupportedPolarity, edgeType.ToString()));
```
NikhilSivadas-NIB, line 1568: Here, we are still referring to ResourceStrings.Digital_TMUUnsupportedPolarity

NikhilSivadas-NIB (commented):
Added a minor comment

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: I think `ValidateAndGetSourceEventForEdge` need to have `ResourceStrings.Digital_TMUUnsupportedPolarity` as we are using `TmuPolarity` as switch cases, so if the given `TmuPolarity` is not supported then we need to throw this error.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Added the exception respective to each of the parameter type.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: Sorry , got confused with parameter name edgetype

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: Good that's added, but other point was if we inherit from overload takes pinNames array input, it will bring parameter pinNames also right? If yes, it will be incorrect for this method right?
@Utkarsh-NI-Emerson can you confirm this?

NikhilSivadas-NIB (commented):
Just one minor comment, can you reply?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: > pinNames array input, it will bring parameter pinNames also right?

No, it won't. The inherited documentation for the pinNames parameter is overridden in the single-pin overloads by explicitly defining:
```csharp
<param name="pinName">The pin name to configure.</param>
```
Because the parameter documentation is explicitly added in these overloads, the `pinNames` parameter documentation is not inherited from the array-based overload. It would only be inherited if we relied entirely on `inheritdoc/>` without providing a replacement `<param>` entry for `pinName`.

From Microsoft Documentation 
> cref: Specify the member to inherit documentation from. The inherited tags _**don't override already defined tags on the current member**_.

So, when we put `<inheritdoc/>` on a member and then explicitly add a `<param name="x">...</param>`  tag for one specific parameter, the compiler inherits everything except that parameter.

Reference : https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#inheritdoc

NikhilSivadas-NIB: Thank you clarification.

NikhilSivadas-NIB (approved):
Approved

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Utkarsh-NI-Emerson: @NikhilSivadas-NIB 
As per my understanding, the explanation from @Ashutosh-NI is correct.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -246,6 +247,645 @@ public void Inititalize_ConfigureTMUSampleTimeoutSucceeds(double timeout, bool u
...
+        }
+
+        #endregion
+
+        #region Configure Period Measurement Tests
+
+        [Theory]
+        [InlineData(TmuPolarity.RisingEdge, 1, false)]
+        [InlineData(TmuPolarity.RisingEdge, 100, true)]
+        public void Inititalize_ConfigurePeriodMeasurementSucceeds(TmuPolarity edgeType, long samplesToAcquire, bool useSpecificPins)
```
Utkarsh-NI-Emerson: @Ashutosh-NI 
Typo in test method names:
"Inititalize" should be "Initialize".
Applicable to all test methods.

Ashutosh-NI: Updated

Ashutosh-NI (commented):
@NikhilSivadas-NIB Following are the updates:
1. Added the parameter to take `TmuArmSetting` instead of `TmuArmType` for arm type and updated `ConfigureAndEnableTmu` to have edge setting related property.
2. Added test to test above changes
3. Updated the documentation of Duty Cycle to have information about -ve measurement and how user can have ratio/percentage.
4. Update pulse width method documentation to have information about -ve values.

Successful test run(newly added) on hardware :

Image

Please have a look at these changes.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -891,9 +891,120 @@ public void Inititalize_TMUInitiateAndAbortWithSinglePinSucceeds()
...
+            sessionsBundle.DisableTMU();
+            sessionsBundle.ClearTMUAssignment();
+        }
+
+        [Fact]
+        public void Initialize_ConfigureTMUPulseWidthMeasurementWithArmSettingAndSinglePinSucceeds()
+        {
+            var sessionsBundle = InitializeAndCreateBundle();
+
+            sessionsBundle.ConfigureTMUPulseWidthMeasurement(TmuPulseWidth.Low, 1, "C0", TmuArmSetting.StartEdge);
```
NikhilSivadas-NIB, line 998: Can you add test cases for "ConfigureTMUSkewMeasurement" ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -1492,6 +1548,14 @@ private static void ValidateArmType(TmuArmType armType)
             }
         }
 
+        private static void ValidateArmSetting(TmuArmSetting armSetting)
```
NikhilSivadas-NIB: can we rename "ValidateTmuArmSetting" to align with enum name?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -891,9 +891,120 @@ public void Inititalize_TMUInitiateAndAbortWithSinglePinSucceeds()
 
         #endregion
 
+        #region TMU Arm Setting Tests
+
+        [Theory]
+        [InlineData(TmuArmSetting.Immediate)]
+        [InlineData(TmuArmSetting.StartEdge)]
+        [InlineData(TmuArmSetting.StopEdge)]
+        public void Initialize_ConfigurePeriodMeasurementWithArmSettingSucceeds(TmuArmSetting armSetting)
```
NikhilSivadas-NIB: we are making sure it does throw exception for each TmuArmSetting , but do we have mechanism to differentiate these tmuArmSetting for each type to ensure that it works?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -891,9 +891,120 @@ public void Inititalize_TMUInitiateAndAbortWithSinglePinSucceeds()
...
+        {
+            var sessionsBundle = InitializeAndCreateBundle();
+
+            sessionsBundle.ConfigurePeriodMeasurement(TmuPolarity.RisingEdge, 1, armSetting);
```
NikhilSivadas-NIB, line 904: can we remove TMUArmType from existing tests names? Also we are going to remove this enum right?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -891,9 +891,120 @@ public void Inititalize_TMUInitiateAndAbortWithSinglePinSucceeds()
...
+            sessionsBundle.DisableTMU();
+            sessionsBundle.ClearTMUAssignment();
+        }
+
+        [Theory]
+        [InlineData(TmuArmSetting.Immediate)]
+        [InlineData(TmuArmSetting.StartEdge)]
+        [InlineData(TmuArmSetting.StopEdge)]
+        public void Initialize_ConfigureTMURiseTimeMeasurementWithArmSettingSucceeds(TmuArmSetting armSetting)
```
NikhilSivadas-NIB: it will be better to split expected behavior with underscore rigt? Same applicable to all tests in this PR?

NikhilSivadas-NIB (commented):
Added few comments, can you check?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: > but do we have mechanism to differentiate these tmuArmSetting for each type to ensure that it works?

To actually verify the working we need hardware and see the value, which we have already done as part of hardware validation work item (as this gap was caught as part of hardware validation).
With that said we are validating that we are able to set correct value for driver as  we are enabling the tmus, also `ClearTMUAssignment` have a driver calls of `GetDisabledTmuContexts` which also internally verifies if all the assigned tmu attribute values are correct.

And proper working of each of edge type is verified in hardware validation.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: >can we remove TMUArmType from existing tests names?

The only test which have `TmuArmType` in its name are the once which validates either of `ConfigureTMUEdgeXX` methods as they still require `TmuArmType`.

>Also we are going to remove this enum right?

 No, as mentioned above we will be still using as input to the `ConfigureTMUEdgeXX` methods and also this enum isuse to set values to different driver attributes.
 
 The introdued `TmuArmSetting` is internal to STL to make the user input more intutive.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: > it will be better to split expected behavior with underscore right?

I have kept the format of the tests here as `Arrange_Act` as there is no `Assert` to these tests, so similar is being reflected in the name of these tests.
let me know your thoughts on this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Added the test for Skew methods

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
NikhilSivadas-NIB: There is no explicit Assert in this test, but the implicit validation is that no exception is thrown right, so it will be easy to read later if its clear in test name

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
NikhilSivadas-NIB: For my understanding, how are we verifying in hardware? is it possible to validate the same here in each type? Right each type does not make any difference in Assert part right.

Ashutosh-NI: > For my understanding, how are we verifying this on hardware?

On hardware, we have a loopback connection that allows us to generate and read the signal. The flow is to initiate the TMU, read the value, and then disable the TMU.

For ATS, the situation is different:

1. We cannot read/fetch a value because there is no actual signal source available. As a result, the operation will always timeout and throw an exception.
2. Even if we skip the fetch/read operation and only initiate and disable the TMU:

```csharp
TMUInitiate();

// Immediately disable
DisableTMU();
```
There are still scenarios where an exception can occur. For example, after TMUInitiate() returns, the driver may still be configuring the edge detection and waiting for an edge signal in the background. If DisableTMU() is called during this state, the operation fails and throw an exception because disabling the TMU is not valid while the TMU is still in the middle of its setup or measurement process.

So, without a valid signal source in ATS, it is difficult to reliably reproduce the same hardware verification flow, and timing-related exceptions are expected in some cases.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Updated the name of the tests.

NikhilSivadas-NIB (approved):
Approved

PR #536, opened 2026-07-07
Create helper method for `Commit` and `Initiate` for attribute update operation

nitamil:

This PR contains a helper method for attribute update operation and changes to Configure Settings APIs and configure current limit APIs and its overloads.

This PR is part of the STL Usabliity improvement and this PR introduces the common helper method that is need for all the upcoming changes. 

Created Autotest to cover all the possible values of the new parameter and tested the same. 
All tests are passing. For commit operation the test logic is same as `Deffered` and the validation of the operation is done with the help of NI IO Trace. Attached the snapshot of IO trace highlighting the commit operation is successfully executed. 

Below image confirms the IO Trace shows that commit is being called as part of the commit option test for each test.
1) Test: DifferentSMUDevices_ConfigureSourceSettingsWithScalarAndUpdateMode_CorrectValuesAreSet.

image

2) Test: DifferentSMUDevices_ConfigureSourceSettingsWithPerSiteAndUpdateMode_CorrectValuesAreSet.
image

3) Test: DifferentSMUDevices_ConfigureSourceSettingsWithPerPinPerSiteAndUpdateMode_CorrectValuesAreSet.
image

4) Test: DifferentSMUDevices_ConfigureSourceSettingsWithPerPinAndUpdateMode_CorrectValuesAreSet.
image

5) Test: DifferentSMUDevices_ConfigureCurrentLimitWithUpdateMode_CorrectValuesAreSet.
image

6) Test: DifferentSMUDevices_ConfigureCurrentLimitsWithUpdateMode_CorrectValuesAreSet.
image


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -58,18 +58,20 @@ private static string BuildSequenceName()
         /// </summary>
         /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
         /// <param name="settings">The source settings to configure.</param>
-        public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, DCPowerSourceSettings settings)
+        /// <param name="updateMode">The update mode to use when applying the settings.</param>
+        public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, DCPowerSourceSettings settings, UpdateMode updateMode = UpdateMode.Deferred)
```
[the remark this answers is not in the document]
nitamil: @Mattjet27, I am making an assumption here that customer has to rebuild thier dlls when using this new bundle with the updated extensions. Will there be any usecases (workflows) which needs the binary compatibility which requires to preserve the old signature.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -116,5 +116,22 @@ public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsB
             }
         }
         #endregion methods on DCPowerSessionsBundle
-    }
+
+        internal static void AttributeUpdate(
```
NikhilSivadas-NIB: can we change the name of method to ApplyUpdateMode or anything more suitable?  AttributeUpdate does not seems to be very matching to what we are doing in the method

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
nitamil: Updated.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -58,18 +58,20 @@ private static string BuildSequenceName()
         /// </summary>
         /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
         /// <param name="settings">The source settings to configure.</param>
-        public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, DCPowerSourceSettings settings)
+        /// <param name="updateMode">The update mode to use when applying the settings.</param>
```
NikhilSivadas-NIB: Spec says inline doc for updateMode parameter "Specifies when the configured settings take effect." do you think that's better? Also, since UpdateMode is a public enum with detailed descriptions for each value, should we reference it in the parameter documentation in the public methods?

NikhilSivadas-NIB (commented):
I have partially reviewed the PR and added few minor comments, can you check it? I will continue review meanwhile

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
nitamil: Yes. That's better. I will updated it with the reference.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
nitamil: `UpdateAttributes` or `UpdateProperties`?

nitamil: Renamed to `UpdatePropeties` as the function is updating the property to the driver.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
Mattjet27: @nitamil, here is my understanding: Maintaining binary compatibility is critical for .NET Framework because of dependent assemblies that are installed on the system and registered to the GAC, which may get upgraded intendent of the test program and redirect to newer versions. However, STL is delivered as a NuGet package, while the Extensions assembly is not registered to the GAC, the Abstraction assembly is. Therefore, it would be necessary to maintain binary compatibility within the Abstractions assembly, for any users that are directly referencing the installed assembly. In all other cases, .NET Framework CLR will load in the version of the assembly that is local to the test program, as is the case for .NET Core as well. NuGet packages will inherently copy the dependent assemblies into the output directory of the compiled program. Assemblies loaded from the application's output directory inherently enforce binary compatibility. Therefore, since users are consuming this code via NuGet package and there is no concern about the dependent assembly getting loaded from the GAC by the CLR, I do not believe we do not strictly need to maintain binary compatible for Extension methods, only source compatibility. 

The one exception to this may be if a user is referencing the Extensions assembly directly from the install location, and that assembly does not get copied into their output directory (i.e. alongside their assembly), but I think that the CLR cannot find it in that case, since it's not in the GAC, so it should not be an issue. However, I am not sure if there would be an issue if multiple versions are loaded be separate applications, for example one version of the assembly by the TestStandSteps, and a different version by the end user's program. We should test this scenario to confirm, but I suspect this is possible and the CLR handles multiple versions loading for such cases, as I think I have seen this done before.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
NikhilSivadas-NIB: UpdateProperties() still feels slightly confusing,  as the properties have already been configured before this method is called, and this method is doing Commit or Initiate the based on UpdateMode, I would suggest "ApplyUpdateMode", also UpdateMode is public Enum right so intent is clear

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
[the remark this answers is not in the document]
nitamil: Yes i have seen multiple versions loaded by TestStand Application as well. We have an explicit API in our library to address the Multiple versions being loaded and the Cache is copied over to the latest Dll. We are good in that case.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -58,18 +58,20 @@ private static string BuildSequenceName()
         /// </summary>
         /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
         /// <param name="settings">The source settings to configure.</param>
-        public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, DCPowerSourceSettings settings)
+        /// <param name="updateMode">The <see cref="UpdateMode"/> object.</param>
```
NikhilSivadas-NIB: Sorry my bad, I just referred inline doc of other public APIS with Enums we exposed via Extensions, it doesn't reference it this way it, can we keep the same inline doc as per spec here for updateMode parameter? "Specifies when the configured settings take effect." Lets follow the same pattern here and other methods too.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -2990,6 +2990,234 @@ public void DifferentSMUDevices_ConfigureSourceSettings_CorrectValuesAreSetWithS
...
+            {
+                Assert.Equal(1.8, sessionInfo.AllChannelsOutput.Source.Voltage.VoltageLevel);
+                Assert.Equal(0.05, sessionInfo.AllChannelsOutput.Source.Voltage.CurrentLimit);
+            });
+            if (updateMode == UpdateMode.Immediate)
+            {
+                var exception = Assert.Throws<NISemiconductorTestException>(InitiateTest);
+                Assert.Contains("The session is already running.", exception.Message);
+            }
+            else
```
NikhilSivadas-NIB, line 3025: is there a way that we could differentiate both UpdateMode.Deferred and UpdateMode.Commit in Assert part? in all of these tests below.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -116,5 +116,23 @@ public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsB
...
+                case UpdateMode.Commit:
+                    sessionsBundle.Commit();
+                    break;
+                case UpdateMode.Immediate:
+                    sessionsBundle.Initiate();
+                    break;
+                case UpdateMode.Deferred:
+                    break;
+                default:
+                    throw new System.ArgumentOutOfRangeException(nameof(updateMode), updateMode, "Unknown update mode.");
```
NikhilSivadas-NIB: Can we change this to throw NISemiconductorTestException and move the message to Resources as well? I think the general practice in Extensions is to encapsulate underlying errors and exceptions in NISemiconductorTestException

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -2990,6 +2990,234 @@ public void DifferentSMUDevices_ConfigureSourceSettings_CorrectValuesAreSetWithS
...
+            {
+                sessionsBundle.Initiate(); // Should not throw exception for Deferred or Commit update modes
+            }
```
NikhilSivadas-NIB, line 3218: Also, as we are explicitly handling invalid UpdateMode, can we add at least one test case for the negative scenario of passing an invalid UpdateMode value?

NikhilSivadas-NIB (commented):
Can you check comments?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: With auto test we cann't differntiate. This is agreed and documented in the Spec. Also I will attach the NI IO trace as part of `Commit` operation execution. After that I will resolve this comment.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
nitamil: We will Default should be same as deferred, i.e. no operation. So removed the exception throwing logic.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: Updated the test result in the Description of the PR.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: Invalid operation is no operation and ignore them.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Control.cs
```diff
@@ -116,5 +116,22 @@ public static void InitiateAdvancedSequence(this DCPowerSessionsBundle sessionsB
...
+                case UpdateMode.Commit:
+                    sessionsBundle.Commit();
+                    break;
+                case UpdateMode.Immediate:
+                    sessionsBundle.Initiate();
+                    break;
+                case UpdateMode.Deferred:
+                default:
+                    break;
+            }
```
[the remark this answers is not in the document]
nitamil: Won't Fix

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
```diff
@@ -2990,6 +2990,234 @@ public void DifferentSMUDevices_ConfigureSourceSettings_CorrectValuesAreSetWithS
             Assert.Equal(settings.SourceDelayInSeconds.Value, sessionsBundle.InstrumentSessions.ElementAt(3).AllChannelsOutput.Source.SourceDelay.TotalSeconds);
         }
 
+        [Theory]
+        [InlineData(UpdateMode.Deferred)]
+        [InlineData(UpdateMode.Commit)]
+        [InlineData(UpdateMode.Immediate)]
+        public void DifferentSMUDevices_ConfigureSourceSettingsWithScalarAndUpdateMode_CorrectValuesAreSet(UpdateMode updateMode)
```
NikhilSivadas-NIB: can we update assert part of test name to "CorrectValuesAreSetAndMatchUpdateMode"? because we are asserting update mode behavior as well by calling initiate , is that ok? same for all tests

docs/UserGuide/advanced/BestPracticesForWritingExtensionMethodsInSTL.md
```diff
@@ -193,7 +193,7 @@ The decision to use Scalar, SiteData, or PinSiteData depends on the level of req
 **Example**:
 
 - ***Input Parameter***: `SiteData` object represents the value for each site.
-  - [public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, SiteData\<DCPowerSourceSettings> settings)](https://github.com/ni/semi-test-library-dotnet/blob/87f9ebe52c1eba721fda454b5c1712bb6bdae77d/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs#L36)
+  - [public static void ConfigureSourceSettings(this DCPowerSessionsBundle sessionsBundle, SiteData\<DCPowerSourceSettings> settings, UpdateMode updateMode = UpdateMode.Deferred)](https://github.com/ni/semi-test-library-dotnet/blob/main/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs#L74)
```
NikhilSivadas-NIB: blob revision has been changed from specific revision(**87f9ebe52c1eba721fda454b5c1712bb6bdae77d**) to **main**, is that intentional? on checking document mostly it refer to specific revision in other areas in the document, so in this case should we refer to this revision which has the latest updates or main? can we confirm?

NikhilSivadas-NIB (commented):
Overall looks ok, added two minor comments

docs/UserGuide/advanced/BestPracticesForWritingExtensionMethodsInSTL.md
nitamil: This is the change recommended by the AI Agent review. we can check with team on this and then we can decide on the next step. 

@Utkarsh-NI-Emerson . Let me know what is the norm we follow for Github documentation?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/SourceTests.cs
nitamil: Sure I will update.

docs/UserGuide/advanced/BestPracticesForWritingExtensionMethodsInSTL.md
nitamil: Moved the changes to next PR.

NikhilSivadas-NIB (approved):
Approved, As discussed please update the signatures in best practices md file based on these changes in separate task post discussion with team.

PR #561, opened 2026-07-21
Updating Assign/Clear TMU Contexts.

Ashutosh-NI:
This pull request refactors the TMU resource assignment and release logic in `TmuExtensions.cs` to improve robustness when handling multiple session bundles and overlapping pin assignments. It also introduces new unit tests to verify these scenarios. The  changes are grouped below:

### TMU Resource Assignment and Release Improvements

* Refactored TMU assignment logic to categorize available TMUs by instrument and assign them using a new `TryGetTMU` helper, replacing previous reliance on a global manager and improving assignment correctness when multiple bundles are involved.
* Updated `ClearAssignedTMUContexts` to accept a `doTMUReleaseCheck` flag and improved safety checks by passing a delegate to retrieve available TMUs, ensuring accurate validation during resource release.
* Changed TMU unassignment to use `TMUContextManager.Instance.UnAssignTMU` for more precise resource management. 

### Test Coverage

* Added a new test region covering assignment and release of TMU resources across multiple bundles, including scenarios with overlapping pins and resource exhaustion, to ensure correct behavior and cleanup.

Work item : [Technical Debt 3820941](https://dev.azure.com/ni/DevCentral/_workitems/edit/3820941): Re-look at TMUContextManager and re-design as needed.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -819,44 +820,39 @@ private static void AssignTMUContexts(this DigitalSessionInformation digitalSess
...
                 // - AssignTMUResources() is invoked after having already invoked AssignTMUResources(pinNames) on the same the bundle object for a subset of pins.
                 // - AssignTMUResources(pinNames1) is invoked after having already invoked AssignTMUResources(pinNames2) on the same the bundle object,
                 // where pinNames1 and pinNames2 contain overlapping pins.
-                if (string.IsNullOrEmpty(digitalSitePinInfo.AssignedTmuContext))
+                if (string.IsNullOrEmpty(assignedTmuContext))
                 {
                     string deviceName = digitalSitePinInfo.InstrumentName;
-                    var success = TMUContextManager.TryCheckOutTMU(deviceName, out string tmuName);
-                    if (!success)
+                    if (!TryGetTMU(tmusPerInstrument, deviceName, out string tmuName))
```
NikhilSivadas-NIB, line 840: method name "TryGetTMU" sounds confusing to me, we are getting and Assigning TMUContext in the below line right?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -819,44 +820,39 @@ private static void AssignTMUContexts(this DigitalSessionInformation digitalSess
 
             // Initialize the TMUAssignmentManager with the available TMU resources for the devices within the current session.
```
NikhilSivadas-NIB, line 822: For my understanding, are we considering this comment?[ClearAssignedTMUOnInitiate](https://dev.azure.com/ni/Users/_git/srd-sebu-stl-tmu-early-access/pullRequest/1232947?discussionId=9592040#1776052195)

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -865,27 +861,28 @@ private static void ClearAssignedTMUContexts(this DigitalSessionInformation digi
             foreach (SitePinInfo sitePinInfo in sitePinInfos)
             {
                 var digitalSitePinInfo = sitePinInfo as DigitalSitePinInfo;
-
+                var assignedTmuContext = digitalSitePinInfo?.AssignedTmuContext;
                 // Clear only if TMU resource is assigned for a site/pin pair.
```
NikhilSivadas-NIB: just to verify, Hope these comments are up to date

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -865,27 +861,28 @@ private static void ClearAssignedTMUContexts(this DigitalSessionInformation digi
...
                 // This can happen when:
                 // - 'ClearTMUAssignment' is invoked twice on the same bundle object.
                 // - 'ClearTMUAssignment' is invoked before invoking 'AssignTMUResources'.
                 // - 'ClearTMUAssignment(pinNames)' is invoked, targeting only a subset of pins within the bundle object, and then the 'ClearTMUAssignment()' is invoked on whole bundle object.
-                if (!string.IsNullOrEmpty(digitalSitePinInfo.AssignedTmuContext))
+                if (!string.IsNullOrEmpty(assignedTmuContext))
                 {
                     string deviceName = digitalSitePinInfo.InstrumentName;
-                    string tmuName = digitalSitePinInfo.AssignedTmuContext;
+                    string tmuName = assignedTmuContext;
```
NikhilSivadas-NIB: do we require this new variables deviceName and tmuName here? digitalSitePinInfo.InstrumentName and assignedTmuContext can be directly passed to TMUContextManager.Instance.UnAssignTMU, have added a comment to consider renaming of UnAssignTMU method in AzDo PR too

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -865,27 +861,28 @@ private static void ClearAssignedTMUContexts(this DigitalSessionInformation digi
...
                     digitalSitePinInfo.AssignedTmuContext = string.Empty;
-                    TMUContextManager.TryCheckInTMU(deviceName, tmuName);
+                    TMUContextManager.Instance.UnAssignTMU(deviceName, tmuName);
                 }
             }
         }
 
-        private static bool IsSafeToReleaseAllTMUs(IEnumerable<SitePinInfo> sitePinInfos, List<string> availableTMUList)
+        private static bool IsSafeToReleaseAllTMUs(IEnumerable<SitePinInfo> sitePinInfos, Func<List<string>> getAvailableTMUs)
```
NikhilSivadas-NIB, line 886: Do we really need this parameter as a delegate? 
Otherwise since both methods are private, can we simplify by passing a List<string> directly?
Alternatively, we could pass digitalSessionInformation and call GetDigitalTmus(digitalSessionInformation.Session).GetDisabledTmuContexts() inside the helper.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -447,6 +447,57 @@ public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsAndEitherEdge_Th
...
+
+            bundle1.AssignTMUResources();
+            bundle2.AssignTMUResources();
+
+            bundle1.ClearTMUAssignment();
+            bundle2.ClearTMUAssignment();
+        }
+
+        [Fact]
+        public void AssignTMUResourcesOnFirstBundle_AssignSamePinOnSecondBundle_Succeeds()
```
NikhilSivadas-NIB, line 467: Can we start the act part of test name with the actual method being tested, sample applicable to other tests

Image

NikhilSivadas-NIB (commented):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -819,44 +820,39 @@ private static void AssignTMUContexts(this DigitalSessionInformation digitalSess
...
+                if (string.IsNullOrEmpty(assignedTmuContext))
                 {
                     string deviceName = digitalSitePinInfo.InstrumentName;
-                    var success = TMUContextManager.TryCheckOutTMU(deviceName, out string tmuName);
-                    if (!success)
+                    if (!TryGetTMU(tmusPerInstrument, deviceName, out string tmuName))
                     {
-                        throw new NISemiconductorTestException(
-                            string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUNotEnoughResources, deviceName, sitePinInfo.SitePinString));
+                        throw new NISemiconductorTestException(string.Format(CultureInfo.InvariantCulture, ResourceStrings.Digital_TMUNotEnoughResources, deviceName));
```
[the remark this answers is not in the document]
Ashutosh-NI: Updated as suggested.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -447,6 +447,57 @@ public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsAndEitherEdge_Th
...
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager();
+            var bundle1 = sessionManager.Digital(new string[] { "C0" });
+            var bundle2 = sessionManager.Digital(new string[] { "C0" });
+            bundle1.AssignTMUResources();
```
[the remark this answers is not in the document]
Ashutosh-NI: Updated the name

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -447,6 +447,57 @@ public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsAndEitherEdge_Th
...
+
+        [Fact]
+        public void AssignTMUResourcesOnFirstBundle_ExhaustsResourcesOnSecondBundle_ThrowsAndCleansUp()
+        {
+            var sessionManager = InitializeSessionsAndCreateSessionManager();
+            var bundle1 = sessionManager.Digital(new string[] { "C0", "C1" });
+            var bundle2 = sessionManager.Digital(new string[] { "C0", "C1" });
+
+            // First bundle claims the available TMU resources.
+            bundle1.AssignTMUResources();
```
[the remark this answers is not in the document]
Ashutosh-NI: Updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using NationalInstruments.ModularInstruments.NIDigital;
```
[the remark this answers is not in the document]
Ashutosh-NI: Removed

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -447,6 +447,57 @@ public void Inititalize_ConfigureSkewMeasurementWithMultiplePinsAndEitherEdge_Th
...
+
+            // Second bundle attempts to claim the same already-assigned resources and should fail.
+            Assert.Throws<NISemiconductorTestException>(() => bundle2.AssignTMUResources());
+
+            // The failed assignment should have rolled back its partial claims,
+            // so releasing the first bundle must still succeed.
+            bundle1.ClearTMUAssignment();
+        }
```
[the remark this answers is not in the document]
Ashutosh-NI: Updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Yes added `ClearAssignedTMUContextsForSitePin` as part of `TMUInititate`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Yes, these are still valid as this are different which are avoid by checking `(!string.IsNullOrEmpty(assignedTmuContext))`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: > we are getting and Assigning TMUContext in the below line right?

@NikhilSivadas-NIB  Will `tryGetTMUForSitePin` be better name ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: I have used delegate here because delegate is avoiding the driver call when we don't need it. At the call site:
```csharp
    if (doTMUReleaseCheck && !IsSafeToReleaseAllTMUs(sitePinInfos, () => GetDigitalTmus(digitalSessionInformation.Session).GetDisabledTmuContexts()))
```
Because of the `&&` short-circuit, when `doTMUReleaseCheck` is `false` the helper is never entered, so `GetDisabledTmuContexts()` is never executed. That matters for the rollback path in `AssignTMUResources`:
```csharp
    catch
    {
        sessionInfo.ClearAssignedTMUContexts(pinNames, doTMUReleaseCheck: false);
        throw;
    }
```
**On passing a `List<string>` directly:** we would have to build the list before the `if`, which means calling into the driver on every rollback, even though the result is discarded. `GetAllDisabledTMUContexts` also calls `CheckStatus()` and can throw, and since this runs inside a `catch`, a new exception there would replace the original one we're trying to rethrow. We could restructure by hoisting the check into its own `if (doTMUReleaseCheck) { ... }` block to keep the current behavior, but that just moves the responsibility to the caller and is behaviorally identical to what we already have.

@NikhilSivadas-NIB Let me know if you have suggestion on this

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Updated the test names

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated the same

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: TryGetTMUContext doesn't suit? i wouls say the out parameter also needs to be renamed to tmuContext to be consistent across, both here and the method parameter in signature, thoughts?

NikhilSivadas-NIB (commented):
can you check this comment, meanwhile i will complete the rest

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: I agree with that point, can we opt for alternative that was suggested ?

`  private static bool IsSafeToReleaseAllTMUs(NIDigital session, IEnumerable<SitePinInfo> sitePinInfos)
  {
      var disabledTmuContexts  = GetDigitalTmus(session).GetDisabledTmuContexts();`

can we simplify this by removing the delegate and passing the NIDigital session directly to the private helper ? that should preserve the current behavior right ?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
NikhilSivadas-NIB: I understand why you chose to keep the name TryGetTMU, Looking through Extension class I see in many places TMUContext is being referred as TMU and where as when it goes to TMUContextManager, its referred as TMUContext, you can keep same `terminology in Extension class if you want to continue with it to keep the consistency within this file, otherwise you will need to change that terminology everywhere in this file.

Only concern is that we it might create some confusion later, but anyway you can decide this.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -1428,6 +1429,38 @@ private static void ConfigureAndEnableTmu(
...
+        {
+            // A null or empty list yields an empty dictionary, so downstream TryGetTMU simply
+            // reports that no TMU resources are available rather than faulting here.
+            if (availableTMUs == null || availableTMUs.Count == 0)
+            {
+                return new Dictionary<string, Queue<string>>();
+            }
+            // Build a dictionary with device name as key and queue of available TMU as value.
+            return availableTMUs.GroupBy(tmu => tmu.Split('/')[0])
+                .ToDictionary(g => g.Key, g => new Queue<string>(g.Select(tmu => tmu)));
```
NikhilSivadas-NIB: is this select iterator required at last? "g.Select(tmu => tmu))" 

can it be like this? 
`
return availableTMUs.GroupBy(tmu => tmu.Split('/')[0])
     .ToDictionary(g => g.Key, g => new Queue<string>(g));
`

NikhilSivadas-NIB (commented):
added 2 more comments

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Thanks for pointing this out! Fixed as per the suggestion.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Fixed as per the suggestion.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Makes sense! I have updated all the instances where I am using `tmuContext` to have the name accordingly as that will also align with the name which we are using in `DigitalSitePinInfo` as `AssignedTmuContext`, adn kept `tmu` only for the places which are return `DigitalTmu` class.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -49,6 +49,8 @@ public static void TMUInitiate(this DigitalSessionsBundle sessionsBundle, string
                     SetDigitalHighZState(sessionInfo);
                     DigitalTmu tmu = GetAssignedTmu(sessionInfo, sitePinInfo);
                     tmu.Initiate();
+                    // After inititaing the TMU we can clear them form our list as then they are no longer "just" soft assigned.
+                    ClearAssignedTMUContextsForSitePin(sitePinInfo);
```
Ashutosh-NI: @NikhilSivadas-NIB Was doing the full workflow test with the changed present in this branch and found out that having `ClearAssignedTMUContextsForSitePin` here will not work as after `TmuInitiate` , user generally call `FetchAveragedTMUMeasurement` this will error out as this will also ty to get the `AssignedTmuContext` but we have already removed it in `TmuInititate`.
So, I think we have to remove it from here.

user work flow code 
```csharp
public static void MeasureDutyCycleWithSTL(ISemiconductorModuleContext tsmContext)
{
    // Configuration parameters for TMU duty cycle measurement.
    int numberOfSamples = 100;           // Number of duty cycle samples to collect.
    double timeoutInSeconds = 5.0;       // Maximum time to wait for measurement completion.

    // Step 1: Query TSM session manager to get the digital sessions bundle associated with the "C0" pin.
    var sessionManager = new TSMSessionManager(tsmContext);
    var digitalPins = sessionManager.Digital("C0");

    // Step 2: (Mandatory) Assign TMU resources to the digital pins.
    // This assigns a TMU resource to each of the pins in the digital sessions bundle object,
    // in this case just the "C0" pin.
    // Note that the TMU hardware resource is not reserved until step 4.
    digitalPins.AssignTMUResources();

    // Step 3: Configure the TMU to perform a high duty cycle measurement.
    // - dutyCycleType: Measure the duration of the high portion of the cycle (rising to falling edge at Voh).
    //   Use TmuDutyCycle.Low to instead measure the duration of the low portion of the cycle.
    // - samplesToAcquire: Number of duty cycle measurements to collect.
    // - armType: Start measurement immediately without waiting for an arm event.
    // Note: This method does NOT enable (reserve) the TMU resource at the hardware level.
    // Note: The returned measurement value is a time duration, not a percentage.
    //       To convert to percentage duty cycle, divide by the signal period.
    digitalPins.ConfigureTMUDutyCycleMeasurement(
        dutyCycleType: TmuDutyCycle.High,
        samplesToAcquire: numberOfSamples,
        armType: TmuArmType.Edge);
    digitalPins.ConfigureTMUEdgeArmSource(pinNames: new[] { "C0" });
    digitalPins.ConfigureTMUEdgeArmSourceEvent(TmuSourceEvent.Voh, pinNames: new[] { "C0" });
    digitalPins.ConfigureTMUEdgeArmPolarity(TmuPolarity.RisingEdge, pinNames: new[] { "C0" });
    // Step 4: Enable (reserve) the TMU resource at the hardware level.
    // This step is required when using ConfigureTMUDutyCycleMeasurement.
    digitalPins.EnableTMU();

    // Step 5: Initiate the TMU measurement.
    digitalPins.TMUInitiate();

    // Step 6: Fetch the averaged measurement results.
    // The TMU collects multiple samples and returns the average high duration in seconds.
    PinSiteData<double> dutyCycleMeasurements = digitalPins.FetchAveragedTMUMeasurement(timeoutInSeconds);
    tsmContext.PublishResults(dutyCycleMeasurements, "res");
    // Step 7: Clean up TMU resources.
    // Always disable the TMU and clear assignments when finished to free up resources.
    digitalPins.DisableTMU();
    digitalPins.ClearTMUAssignment();
}
```

Ashutosh-NI: @NikhilSivadas-NIB Removed `ClearAssignedTMUContextsForSitePin ` from `TmuInititate`

NikhilSivadas-NIB (approved):
Approved

PR #573, opened 2026-07-27
Adding Utility Extension Methods for FGen

hcnagaraj-e:

1. Added Extension methods for configure `Reset` and `ResetDevice`
2. Added 4-unit tests to cover testing of these Extension methods
3. Added FGen dependency in Extensions Project and Extension Test project
4. Added 4 PinMap files for running unit tests in offline mode (will be submitted as part of another PR)

Extension methods are added as part of usability improvements.

Feature: [ [STL] [New Instrument Support] [FGen] Support Standard Waveform Generation](https://dev.azure.com/ni/DevCentral/_workitems/edit/2631247)

Feature Spec: [FGen Standard Waveform Generation Support](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/165216/FGen-Standard-Waveform-Generation-Support)

Work Item: [Implement Utility Extensions for FGen] (https://dev.azure.com/ni/DevCentral/_workitems/edit/2631247)

- [x] Tested running unit tests on Offline mode as well as with real hardware
- [x] Ensured general guidelines are followed
- [x] Tested for both single pin and multiple pins cases

Integration tests will be added as part of another workitem once implementation of all the other extension methods are completed.
[Create HW Setup and add additional Integration tests and validate it](https://dev.azure.com/ni/DevCentral/_workitems/edit/3861495)


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGen
+{
+    /// <summary>
+    /// Defines utility methods for NI-FGen session.
+    /// </summary>
+    public static class Utilities
+    {
+        /// <summary>
+        /// Resets the instrument to a known state. This VI aborts the generation, clears all routes, and resets session properties to the default values.
```
gmrsekar:
```suggestion
        /// Resets the instrument to a known state. This method aborts the generation, clears all routes, and resets session properties to the default values.
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,37 @@
+﻿using NationalInstruments.SemiconductorTestLibrary.Common;
+using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.Fgen;
+
+namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGen
```
gmrsekar: Namespace should be same as the one in Abstractions. `NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.Fgen`

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction;
+using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.Fgen;
+using NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.FGen;
```
gmrsekar: This would only be one line after keeping the namespace in Extensions same as in Abstractions.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,60 @@
...
+using NationalInstruments.TestStand.SemiconductorModule.CodeModuleAPI;
+using Xunit;
+using static NationalInstruments.Tests.SemiconductorTestLibrary.Utilities.TSMContext;
+
+namespace NationalInstruments.Tests.SemiconductorTestLibrary.Unit.InstrumentAbstraction.FGen
+{
+    [Collection("NonParallelizable")]
+    public sealed class Utilities : IDisposable
```
gmrsekar: Should it be UtilitiesTests? Usually for test classes, we add the suffix "tests" to the class name.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: Updated to method, copied documentation from DotNet methods

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: Changed to Fgen

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: Yes, only one entry is there now.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: Renamed to UtilitiesTests.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,36 @@
+﻿using NationalInstruments.SemiconductorTestLibrary.Common;
```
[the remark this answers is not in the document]
hcnagaraj-e: It is being used. `Do` methods are taken from that

TestAssets/Supporting Materials/Pin Maps/FgenSingleInstrumentPerPin.pinmap
```diff
@@ -0,0 +1,39 @@
...
+		<DUTPin name="D" />
+		<SystemPin name="SystemPin" />
+	</Pins>
+	<PinGroups>
+		<PinGroup name="PinGroup">
+			<PinReference pin="A" />
+			<PinReference pin="C" />
+			<PinReference pin="SystemPin" />
+		</PinGroup>
+	</PinGroups>
```
Ashutosh-NI: Can we use the pin names similar to `VCC` or `VDD`, as that will be similar to what has been used in other pin maps?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void Dispose()
+        {
+            InitializeAndClose.Close(_tsmContext);
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformResetOperation_Succeeds(string pinmap)
```
Ashutosh-NI: The name of the test and body of the test does not align with each other as the name contains Arrange_Act_Assert but body only have Arrange/Act_Assert. Can we please name it accordingly?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <remarks>
+        /// This method aborts signal generation, resets all attributes to default values, and stops the export of all external signals and events.
+        /// </remarks>
+        public static void Reset(this FgenSessionsBundle sessionsBundle)
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.Utility.Reset();
+            });
+        }
```
Ashutosh-NI, line 23: Why we are having the `Reset` as an extension on `FgenSessionsBundle`  for Fgen and not similar to [`Reset`](https://dev.azure.com/ni/DevCentral/_git/srd-sebu-mixed-signal-apt?path=/SemiconductorTestLibraryDotNET/SemiconductorTestLibraryAbstractions/source/InstrumentAbstraction/DCPower/InitializeAndClose.cs&version=GBusers/usaxena/azdo_TMU_GetterMethods&line=51&lineEnd=69&lineStartColumn=9&lineEndColumn=10&lineStyle=plain&_a=contents) in DCPower or other devices?

Ashutosh-NI (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: We need to add it to DC Power as well, we missed it initially. Matt specifically asked for these two APIs. It is acknowledged in the feature spec.

TestAssets/Supporting Materials/Pin Maps/FgenSingleInstrumentPerPin.pinmap
hcnagaraj-e: VCC VDD is not applicable for FGen use case. Max we can refer as Signal, Signal B. or AnalogInput 1, A01, etc.
For unit test, I feel it's better use generic names. As we do not have full context in the unit test.

I see there are pinmaps using A, B, C and D as pin names.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: I have removed 2nd underscore and made it Arrange_ActAssert format

gmrsekar: Can we just use `Fact` and pass the pin map file name directly to `Intialize` call in line 31 since only one pin map is used?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.Reset();
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
```
gmrsekar, line 48: Similarly, can we just use `Fact` here?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: I was not using fact for two reasons. 
1. We my need to add more pinmap files if needed in future.
2. If We are OK with not calling bundle overload which accepts scalar pin name. Then I can pass pinnames array as test inputs

hcnagaraj-e: Alos, it was more readable to know which pinmaps are being used.

hcnagaraj-e: I have changed to [Fact] for now

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: Changed to `Fact`

gmrsekar (approved):

Ashutosh-NI (approved):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformResetOperationSucceeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.Reset();
```
NikhilSivadas-NIB, line 40: is it possible to ensure state or session properties to ensure its reset?

API reference says, it aborts generation, clears routes, and resets session properties to their defaults

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformResetDeviceOperationSucceeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.ResetDevice();
```
NikhilSivadas-NIB, line 58: same comment as previous comment

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        public void InitializeBundleWithMultiplePins_PerformResetOperationSucceeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen(new string[] { "A", "B", "SystemPin" });
+
+            sessionsBundle.Reset();
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
```
NikhilSivadas-NIB, line 86: To clarify, looks tests focused on single instrument, do we need to consider some tests with multiple instruments ?

NikhilSivadas-NIB (commented):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: Pinmap covers multiple instruments, but mapping of two channels of an instrument across sites and pins are different in each pinmap. 
1. SingleInstrumentPerPin refers to the case where both channels of an instrument map to a single instruments distributed across site-0 and site-1.
2.SingleInstrumentPerSite is that site-0 of pin A and Pin B.
Multiple pins bundle will cover multiple instruments.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: There is no direct way to verify that. It needs fully functional program to evaluate which needs other extension methods being implemented in other PRs.
We have a workitem to add integration tests to cover some of this use case.

For unit test, we can add mock class, or we can use IO Trace to track it.

I have attached the screenshot of the IO trace. Also, added trace.txt as attachment.

image

[UtilityTest IO trace.txt](https://github.com/user-attachments/files/31176070/UtilityTest.IO.trace.txt)

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: I have attached screenshot of the IO trace.
image
[UtilityTest IO trace.txt](https://github.com/user-attachments/files/31176053/UtilityTest.IO.trace.txt)

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.Reset();
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformResetDeviceOperationSucceeds(string pinmap)
```
NikhilSivadas-NIB: can we rename Act part of all tests to public API name? For Example: InitializeBundleWithSinglePin_ResetDeviceSucceeds

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            InitializeAndClose.Close(_tsmContext);
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformResetOperationSucceeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
```
NikhilSivadas-NIB, line 35: Looks other pin maps tend to use pin names closer to real use cases, is it possible to have similar pin names here as well instead of A etc? or is that valid for GFen?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
NikhilSivadas-NIB: but as far as unit tests concerned, can we set some property to non default value as part of arrange and call reset, then assert those properties with default values?  i was trying to find online, can we use output mode or any such properties?

NikhilSivadas-NIB (commented):
added few comments can you check?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: @NikhilSivadas-NIB Act part of the test already has public API name in it "PerformResetDeviceOperation", Do you want to change it to just "ResetDevice" ?

NikhilSivadas-NIB: yes, like this Arrange_ResetDeviceXX_Assert, XX could any small description if has something, otherwise just the method name, Also plan to separate Assert part in test name if we address this comment https://github.com/ni/semi-test-library-dotnet/pull/573/changes#r3818846369.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: Properties do not get reset to default after `Reset` or `Reset` operation is performed. Only things like calibration data, waveform data are being reset. As part of This feature, we are not touching any of those area.
Other than that waveform generation is aborted. 
We need to do multiple settings and start generation and then we need to call `Reset` or `ResetDevice`. Then the generation is aborted. Even then we can't differentiate effect of `Reset` and `Abort` operations.

OutputMode does not get reset on calling 'Reset' or `ResetDevice` driver method. I have verified this behavour.

As I mentioned in the description, we have planned separate work items to add integration tests. We can test multiple workflows once all the extension methods are completed.

If we really need to validate at unit test level, we need to add Mock class to validate the calls.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: Pin names like VCC, VDD are not applicable here. If we see from DUT perspective, we can probably use AO0, AO2, something like that we can use. (Only Analog Output channels can be used). 

For unit tests, we cover mostly the on how each channel of the FGen instrument are being mapped to different pins.
Using pin names like AO0, AO2 will not add clarity in unit tests. Also, existing pinmap for FPGA are also using  pin names as A, B, C, D.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
NikhilSivadas-NIB: @lornazh @Mattjet27 The current unit tests does not have an assertion. We were trying to find a way to verify that the Reset call was executed successfully. My understanding from the API reference is that session properties should be restored to their default value on Reset, but this does not appear to be feasible be verifiable with the current unit test setup, as mentioned by @hcnagaraj-e . 

- is there any other way we could assert the reset call ? or is verifying that no exception is thrown sufficient for this scenario ?

NikhilSivadas-NIB (commented):
can you check the comments? I have tagged Matt and Lorna as well to confirm if this is ok to proceed without Assert.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
Mattjet27: > _"Properties do not get reset to default after `Reset` or `Reset` operation is performed."_

@hcnagaraj-e, I don't think this statement is true. When `Reset` is called, all driver properties should be reset to their default values across the board. The driver API reference for the `Reset` method documentation clearly states, _"This method aborts signal generation, **resets all attributes to default values**, and stops the export of all external signals and events."_. Attributes means properties in this context. 

> `OutputMode` does not get reset on calling `Reset` or ResetDevice driver method. I have verified this behavour

How have you verified this? Was it with actual hardware, or simulated hardware in offline mode? This may not be supported correctly in Offline Mode, FYI.

---

@NikhilSivadas-NIB, it should be possible to cache the initial state of a driver property, change its state and commit, then verify it changed, reset, and then check its state again, to verify it was reset.

```cs
var sessionManager = Initialize(pinmap);
var sessionsBundle = sessionManager.Fgen("A");
// Cache Initial State (Default)
var propValDefault = sessionsBundle2.DoAndReturnPerInstrumentPerChannelResults(x => x.Session.Output.OutputMode);
// Set Value, Commit, and Verify New Value
sessionsBundle2.DoAndReturnPerInstrumentPerChannelResults(x => x.Session.Output.OutputMode = ModularInstruments.NIFgen.OutputMode.Sequence);
var propValAfterMod = sessionsBundle2.DoAndReturnPerInstrumentPerChannelResults(x => x.Session.Output.OutputMode);
Assert.NotEqual(propValAfterMod, propValDefault);
// Reset & Verify Default Value
sessionsBundle2.Reset();
var propValAfterReset = sessionsBundle2.DoAndReturnPerInstrumentPerChannelResults(x => x.Session.Output.OutputMode);
Assert.Equal(propValAfterReset, propValDefault);
```

It should also be possible via driver properties buffer (`Utility.ExportAttributeConfigurationBuffer` and `Utility.ImportAttributeConfigurationBuffer`), if you wanted to verify the whole session state, not just one property.

NikhilSivadas-NIB: Thank you @Mattjet27 for detailed response. @hcnagaraj-e Hope we could follow this approach in all the tests here?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
NikhilSivadas-NIB: Ok, But It would be better if we could name it like PinA,  PinB etc instead of A, B if you think real or similar pin names (AO0, AO2) confuse user.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Performs a hard reset on the device. Generation is stopped, all routes are released, external bidirectional terminals are tri-stated,
+        /// FPGAs are reset, hardware is configured to its default state, and all session properties are reset to their default states.
```
NikhilSivadas-NIB, line 27: .NET API reference says Reset or ResetDevice could throw IviCDriveException and other , Do method would wrap it in STL exception, So can we document this in API inline document ?

Image

NikhilSivadas-NIB (commented):
Can you have a look?

NikhilSivadas-NIB (commented):
Updated comment

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: @NikhilSivadas-NIB I went through existing pinmap files, I did not see any pinmap with pin names as 'PinA', `PinB`, etc.
But we do have existing pin maps with pin names as 'A', 'B', etc. Its being used for Custom Instrument and Pre-existing FGen Pinmap.

If we want to change to PinA then probably we need to update all existing PinMaps of Fgen. 

Can we take that as separate PR? Now, we have pinmap file already submitted and all PRs are using that PinMap. 
I can update all the tests at once after the Pinmap is updated.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: @NikhilSivadas-NIB  "ObjectDisposedException" Exception and "IVICDriverException" will be there for all FGen calls, it will be thrown when object session/reference is invalid. Do we need to call them out in inline documentation? 
I have been adding exception tag only for the once STL is generating. 

I will check DCPower implementation and see how they have handled it. I will update accordingly, if other driver implementation is missing them then we can check with Lorna or Matt.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: @Mattjet27 @NikhilSivadas-NIB Its working with actual hardware. I have updated the assertion part, will push the code once tested properly.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: I have updated the name to use just method name for action part and added assert part as well.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: @NikhilSivadas-NIB I have updated the assert part based on output mode

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: @NikilShivadas-NIB I checked DCPower implementation, there are also, these kind of exceptions are not called out in inline documentation.

DCpower document

Image


DCpower extension implementation

Image

NikhilSivadas-NIB: Its better to document as the underlying API clearly documented it. We are invoking Do method here, so it would wrap it into our STLException if IVICDriverException or any other exception occurred so its better to document.
"The underlying NI-FGEN driver returned an error. "

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
NikhilSivadas-NIB: Sure, that's fine

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            var outputModeDefault = sessionsBundle.DoAndReturnPerInstrumentPerChannelResults(sessionInfo => sessionInfo.Session.Output.OutputMode);
+            sessionsBundle.DoAndReturnPerInstrumentPerChannelResults(sessionInfo => sessionInfo.Session.Output.OutputMode = OutputMode.Sequence);
+            var outputModeNonDefault = sessionsBundle.DoAndReturnPerInstrumentPerChannelResults(sessionInfo => sessionInfo.Session.Output.OutputMode);
+
+            sessionsBundle.Reset();
+
+            var outputModeDefaultAfterReset = sessionsBundle.DoAndReturnPerInstrumentPerChannelResults(sessionInfo => sessionInfo.Session.Output.OutputMode);
+            Assert.NotEqual(outputModeNonDefault, outputModeDefault);
+            Assert.Equal(outputModeDefaultAfterReset, outputModeDefault);
```
NikhilSivadas-NIB: can we correct the order of arguments in each Asserts, Expected then Actual ?


Assert.NotEqual(outputModeDefault, outputModeNonDefault);
Assert.Equal(outputModeDefault, outputModeDefaultAfterReset);

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public void Dispose()
+        {
+            InitializeAndClose.Close(_tsmContext);
+        }
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
```
NikhilSivadas-NIB, line 31: Asking for my understanding as well, why are we not considering the other 2 pinmap file for single Pin Bundle tests?

NikhilSivadas-NIB (commented):
can you check?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: Sure, I got confused. I will update

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: When we make bundle for single pin, for other two pinmap files, we end up getting bundle where only one channel of an FGen instrument will be part of the bundle. So, effectively it falls in unsupported bundle case. Right now, it does not fail. But once we add validation logic, it throws exception at bundle creation time.

In other pinmap, instrument channels are mapped to two different pins. Unless we create bundle with both pins, both channels of FGen will not be in bundle. We are not supporting that use case.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/UtilitiesTests.cs
hcnagaraj-e: I have made the correction and tested it again, it is working fine.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Utilities.cs
hcnagaraj-e: Every driver API will have this expectation as they would throw exception if session is invalid. 
I checked DC Power, there also, they are documenting exceptions only when STL throws exception except STL wrapped exceptions for driver exceptions. DO want to still Add Exception tag? I can add it in remarks if that makes sense.

Then we need to update every extension method as it's a generic exception from driver.

NikhilSivadas-NIB: ok, i see couple of Digital method documented IVICDriverException inline, but as you mentioned majority of the methods have not documented this. I think we can go ahead without this for now

NikhilSivadas-NIB (approved):
Approved

PR #574, opened 2026-07-29
[Example] - Creating Copies of Existing Examples and Replacing the Implementation with New Methods

gmrsekar:

This pull request adds a series of new example methods to the `SiteDataExamples.cs` and `PinSiteDataExamples.cs` files. These methods demonstrate how to construct and populate `SiteData<T>` and `PinSiteData<T>` objects dynamically using their empty constructors, and how to add pins, sites, and set values in various scenarios (including site-agnostic/system pins). The examples cover a wide range of use cases, such as per-pin, per-site, per-pin-per-site, and site-agnostic data.

The most important changes are:

**New dynamic construction examples for PinSiteData:**
* Added methods such as `BuildWithWithArraysAndSetValueWithPerPinData`, `BuildWithArraysAndSetValueWithPerPinPerSiteData`, `BuildWithArraysWithSystemPin`, `BuildWithDictionaryWithSystemPin`, `BuildWithPinDataDictionaryAndSiteNumbersArray`, `BuildWithPinDataDictionaryAndSiteNumbersArrayWithSystemPin`, `BuildWithArraysForCommonDataValue`, `BuildWithPinUniqueDataArray`, `BuildWithSiteUniqueDataArray`, `BuildWithPinAndSiteUniqueDataArray`, and `BuildWithSiteAndPinUniqueDataArray` to show how to construct and fill `PinSiteData<double>` objects dynamically in a variety of advanced scenarios. 

**New dynamic construction examples for SiteData:**
* Added methods such as `BuildWithArray`, `BuildWithPerSiteDataDictionary`, `BuildWithDictionaryWithSystemData`, and `BuildWithSingleValue` to show how to build and populate `SiteData<double>` objects using the empty constructor, with explicit site addition and value assignment, including handling site-agnostic data.

**System pin and site-agnostic data handling:**
* Several examples now explicitly demonstrate how to handle system pins (using site number -1) and site-agnostic values, clarifying best practices for these special cases.

These additions provide comprehensive guidance for users who need to construct `SiteData` and `PinSiteData` objects programmatically, especially in cases where pin and site information is not known upfront or when working with system pins and site-agnostic data.

Copies of existing examples are created using new methods to illustrate how we can use the same per-pin, per-site and other types of inputs to build the PinSiteData/SiteData objects dynamically using the new methods.
This PR is created as part of the user story [[5.2] Update Existing Code Snippet Examples with a Copy of Existing Implementation That Uses New Methods](https://dev.azure.com/ni/DevCentral/_workitems/edit/3928943)

Added unit tests that runs all the new examples added.


Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -160,6 +160,49 @@ internal static void ConstructWithArrays()
...
+            }
+        }
+
+        internal static void BuildWithArraysAndSetValueWithPerPinPerSiteData()
+        {
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VCC1", "VCC2" };
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 0, 1, 2, 3 };
+            // Per-pin SiteData objects.
```
Mattjet27: @gmrsekar, this is a copy-and-paste error. The comment should read:

```cs
// 2D jagged array of pin and site unique data,
// where the first dimension represents pins (2) and the second dimension represents sites (4). 
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -180,6 +223,33 @@ internal static void ConstructWithArraysWithSystemPin()
             var pinSiteData = new PinSiteData<double>(pinNames, perPinSiteData);
         }
 
+        internal static void BuildWithArraysWithSystemPin()
```
Mattjet27, line 227: @gmrsekar, this is not what I was expecting to see, moreover we need to ensure the System Pin case is well enough tested and validated with the new methods. Here is what I had in mind (take note of my comments... I think we need to update some of the `SetValue` overload remarks to discourage use of certain overloads when system pins are present):

```cs
internal static void BuildWithArraysWithSystemPin()
{
    // Pin names to associate with the data.
    var pinNames = new string[] { "VCC1", "VCC2", "SystemSupply" };
    // Site numbers to associate with the DUT pins.
    var siteNumbers = new int[] { 0, 1 };
    // Per-pin, per-site data array.
    // Note that data associated with system pins is considered site-agnostic,
    // and site-agnostic data can represented with -1 as the site value.
    var perPinData = new double[] { 1.5, 2.5, -22.5 };

    // Since both pin names and site numbers are known, providing this information to the constructor is most efficient.
    // Alternatively, you can create empty PinSiteData and then add pins and sites manually with the AddPin and AddSite method,
    // or have them be added dynamically as specified by the SetValue method.
    // Note that the constructor's siteNumbers input parameter can accept both an array of site numbers and a single site number,
    // In this example the -1 site value is simply appended after the siteNumbers array,
    // where the -1 value represents the site-agnostic data for any system pins.
    var pinSiteData = new PinSiteData<double>(pinNames: pinNames, siteNumbers: siteNumbers, -1);
    for (int i = 0; i < perPinData.Length; i++)
    {
        // Check if the pin is a system pin and set the value accordingly.
        // In this example, we are assuming that the system pin is named "SystemSupply".
        // However, in practice, you would want to have a more robust way to identify system pins,
        // such as utilizing the ISemiconductorModuleContext.GetPins method.
        if (pinNames[i] == "SystemSupply")
        {
            // For system pins, set the value for site -1 to indicate site-agnostic data.
            pinSiteData.SetValue(value: perPinData[i], pinName: pinNames[i], siteNumbers: -1);
        }
        else
        {
            // Set the uniform value for each DUT pin across all sites.
            // Note that care must be taken when calling SetValue with system pins present in the PinSiteData object,
            // As the  SetValue(T value) and SetValue(T value, param string) overload will apply the same value to all sites declared within the PinSiteData, including system sites. 
            // Therefore, it is best to avoid those overloads when working with system pins,
            // and instead, explicitly specify which site number to set a value for, as shown below. 
            pinSiteData.SetValue(value: perPinData[i], pinName: pinNames[i], siteNumbers: siteNumbers);
        }
    }
}
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -205,6 +302,29 @@ internal static void ConstructWithPinDataDictionaryAndSiteNumbersArray()
...
+        {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 3, 1 };
+            // Dictionary containing pin-unique data.
+            var perPinData = new Dictionary<string, double> { ["VDET"] = 22, ["VCC1"] = 44, ["VCC2"] = 33 };
+            // Get the pin names from the dictionary keys.
+            var pinNames = perPinData.Keys.ToArray();
+            // Use the empty constructor to build the PinSiteData dynamically.
+            // This is useful when pin names or site numbers are not all known upfront.
+            // For this case, we know the pin names and site numbers, so we could also use the constructor that takes both as input.
```
Mattjet27: _For this example, since the pin names and site numbers are known, it would also be possible to provide that information directly to the constructor._

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -249,6 +405,28 @@ internal static void ConstructWithPinUniqueDataArray()
...
+        {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 3 };
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VDET", "VCC1" };
+            // Per-pin data values, where each element is the value for the pin at the same index in pinNames.
+            var perPinData = new double[] { 42, 105 };
+            // Use the empty constructor to build the PinSiteData dynamically.
+            // This is useful when pin names or site numbers are not all known upfront.
+            // For this case, we know the pin names and site numbers, so we could also use the constructor that takes both as input.
```
Mattjet27: _For this example, since the pin names and site numbers are known, it would also be possible to provide that information directly to the constructor._

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -222,6 +342,29 @@ internal static void ConstructWithPinDataDictionaryAndSiteNumbersArrayWithSystem
...
+            var pinSiteData = new PinSiteData<double>();
+            // Set the per-pin value for each DUT pin, repeated across all sites for that pin.
+            // All the pins and sites are added dynamically to the PinSiteData object as they are encountered.
+            foreach (var pinName in dutPinNames)
+            {
+                pinSiteData.SetValue(perDutPinData[pinName], pinName, siteNumbers);
+            }
+            // System pins are site-agnostic; -1 is used as the site number.
+            // SetValue automatically adds the SystemSupply pin and site -1 since they do not yet exist.
+            pinSiteData.SetValue(-15, "SystemSupply", -1);
```
Mattjet27: @gmrsekar, it would be better practice to also have this be in a loop that operates over `perSystemPinSiteData.Keys`. Note that you should not need to cast this to an array if all we are doing is looping over it. The same applies to the `var dutPinNames = perDutPinData.Key;`

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -180,6 +223,33 @@ internal static void ConstructWithArraysWithSystemPin()
...
+            // Per-pin SiteData objects.
+            // Note that data associated with system pins is considered site-agnostic,
+            // and site-agnostic data can represented with -1 as the site value.
+            var perPinSiteData = new[]
+            {
+                new SiteData<double>(siteNumbers, 1.5),
+                new SiteData<double>(siteNumbers, 2.5),
+                new SiteData<double>(new[] { -1 }, -22.5)
+            };
+            // Since both pin names and site numbers are known, using this constructor is more efficient.
```
Mattjet27: @gmrsekar, please update this to read:
```cs
// Since both pin names and site numbers are known, providing this information to the constructor is most efficient.
// Alternatively, you can create empty PinSiteData and then add pins and sites manually with the AddPin and AddSite method,
// or have them be added dynamically as specified by the SetValue method.
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -232,6 +375,19 @@ internal static void ConstructWithArraysForCommonDataValue()
             var pinSiteData = new PinSiteData<double>(pinNames, siteNumbers, 55);
         }
 
+        internal static void BuildWithArraysForCommonDataValue()
+        {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 3 };
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VDET", "VCC1", "VCC2" };
+            // Since both pin names and site numbers are known, using this constructor is more efficient.
```
Mattjet27: @gmrsekar, please update this to read:
```cs
// Since both pin names and site numbers are known, providing this information to the constructor is most efficient.
// Alternatively, you can create empty PinSiteData and then add pins and sites manually with the AddPin and AddSite method,
// or have them be added dynamically as specified by the SetValue method.
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -266,6 +444,24 @@ internal static void ConstructWithSiteUniqueDataArray()
...
 
+        internal static void BuildWithSiteUniqueDataArray()
+        {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 3 };
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VDET", "VCC1" };
+            // Per-site data values, where each element is the value for the site at the same index in siteNumbers.
+            var perSiteData = new double[] { 42, 105, 55 };
+            // Since both pin names and site numbers are known, using this constructor is more efficient.
```
Mattjet27: @gmrsekar, please update this to read:
```cs
// Since both pin names and site numbers are known, providing this information to the constructor is most efficient.
// Alternatively, you can create empty PinSiteData and then add pins and sites manually with the AddPin and AddSite method,
// or have them be added dynamically as specified by the SetValue method.
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -310,6 +534,34 @@ internal static void ConstructWithSiteAndPinUniqueDataArray()
...
+            var pinNames = new string[] { "VDET", "VCC1" };
+            // 2D jagged array of pin and site unique data,
+            // where the first dimension represents sites (3) and the second dimension represents pins (2).
+            var perSitePerPinData = new double[][]
+            {
+                new double[] { 42,  55 },
+                new double[] { 105, 2048 },
+                new double[] { 206, 0.5 },
+            };
+            // Since both pin names and site numbers are known, using this constructor is more efficient.
```
Mattjet27: @gmrsekar, please update this to read:
```cs
// Since both pin names and site numbers are known, providing this information to the constructor is most efficient.
// Alternatively, you can create empty PinSiteData and then add pins and sites manually with the AddPin and AddSite method,
// or have them be added dynamically as specified by the SetValue method.
```

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -288,6 +484,34 @@ internal static void ConstructWithPinAndSiteUniqueDataArray()
...
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VDET", "VCC1" };
+            // 2D jagged array of pin and site unique data,
+            // where the first dimension represents pins (2) and the second dimension represents sites (3).
+            var perPinPerSiteData = new double[][]
+            {
+                new double[] { 42, 105, 206 },
+                new double[] { 55, 2048, 0.5 }
+            };
+            // Start with a single pin using the single-pin constructor, then add the remaining pins.
```
Mattjet27: @gmrsekar, I feel this implementation is slightly more confusing than it is valuable, consider reworking it as the following, so that we demonstrate something more meaningful like the pin-only constructor. 
```cs
// 2D jagged array of pin and site unique data,
// where the first dimension represents pins (2) and the second dimension represents sites (3).
var perPinPerSiteData = new double[][]
{
    new double[] { 42, 105, 206 },
    new double[] { 55, 2048, 0.5 }
};

// Use the pin-only constructor, then add the sites.
// This constructor allows you to declare pins upfront and add sites later on.
var pinSiteData = new PinSiteData<double>(pinNames: "VDET", "VCC1");
// Add sites
pinSiteData.AddSite(2, 4, 3);
// Set a unique value for each pin and each site combination.
for (int pinIndex = 0; pinIndex < pinSiteData.PinNames.Length; pinIndex++)
{
    string pinName = pinSiteData.PinNames[pinIndex];
    for (int siteIndex = 0; siteIndex < pinSiteData.SiteNumbers.Length; siteIndex++)
    {
        int siteNumber = pinSiteData.SiteNumbers[pinIndex]
        pinSiteData.SetValue(perPinPerSiteData[pinIndex][siteIndex], pinName, siteNumber);
    }
}
```

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -133,12 +133,44 @@ internal static void ConstructWithArray()
             // For example the active sites could be: 4, 9, 12, 32.
             var siteData = new SiteData<double>(new double[] { 1, 2, 3 });
         }
+
+        internal static void BuildWithArray()
+        {
+            // The data values to set per site, where the index represents the site number.
+            var data = new double[] { 1, 2, 3 };
+            // Use the empty constructor to build the SiteData dynamically.
+            // This allows sites to be added and values to be set individually,
```
Mattjet27: @gmrsekar, you do not need these two comments on lines 142 and 143, please remove them. Note that they are redundant based on the first comment made on line 139.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -133,12 +133,44 @@ internal static void ConstructWithArray()
...
+            // rather than relying on array-index-to-site-number mapping.
+            var siteData = new SiteData<double>();
+            // Add site numbers explicitly so the mapping is clear and not index-dependent.
```
Mattjet27: @gmrsekar, please change this to read, _"Add site numbers sequentially from zero, since the index of the data above represents the site number."_.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -133,12 +133,44 @@ internal static void ConstructWithArray()
...
         }
+
+        internal static void BuildWithPerSiteDataDictionary()
+        {
+            // Dictionary containing site-unique data values.
+            var perSiteDataDictionary = new Dictionary<int, double> { [1] = 11, [2] = 22, [3] = 33 };
+            // Use the empty constructor to build the SiteData dynamically.
+            var siteData = new SiteData<double>();
+            // Set each value for the corresponding site number individually. Site numbers are added automatically if they do not already exist in the SiteData object.
+            foreach (var entry in perSiteDataDictionary)
```
Mattjet27: @gmrsekar, it's more typically to use "item" instead of "entry" when referring to a dictionary. Consider changing this.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -176,6 +240,25 @@ internal static void ConstructWithSiteUniqueDataArray()
...
+            var siteNumbers = new int[] { 2, 4, 3 };
+            // Array of site unique data, where each element represents the data for the specific site,
+            // located at the same index within the siteNumbers array.
+            var perSiteData = new double[] { 22, 44, 33 };
+            // Start with the single-site constructor for the first site, then add the remaining sites.
+            var siteData = new SiteData<double>(siteNumber: siteNumbers[0]);
+            // Set the value for the first site.
+            siteData.SetValue(perSiteData[0], siteNumbers[0]);
+            // Add remaining site numbers and set their corresponding values individually.
+            // Non-sequential site order is accepted.
```
Mattjet27, line 250: @gmrsekar, I am not sure you need to set the first site separately. Consider reworking this to just have all values set via the loop.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -158,6 +207,21 @@ internal static void ConstructWithSingleValue()
...
             var siteData = new SiteData<double>(siteNumbers, data);
         }
+
+        internal static void BuildWithSingleValue()
+        {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 5, 6 };
+            var data = 567;
+            // Use the empty constructor to build the SiteData dynamically.
+            var siteData = new SiteData<double>();
```
Mattjet27, line 214: @gmrsekar, in this case it might be more straight forward to just use the constructor approach, and also not relay on needing any variable, the example can simply be as follows:
```cs
// Create a new SiteData object for the following sites.
var siteData = new SiteData<double>(siteNumbers: 2, 4, 5, 6);
// Set the same value across all sites at once.
siteData.SetValue(value: 567);
```

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
```diff
@@ -147,6 +179,23 @@ internal static void ConstructWithDictionaryWithSystemData()
             var perSiteDataDictionary = new Dictionary<int, double> { [1] = 11, [2] = 22, [-1] = 33 };
             var siteData = new SiteData<double>(perSiteDataDictionary);
         }
+
+        internal static void BuildWithDictionaryWithSystemData()
```
Mattjet27, line 181: @gmrsekar, please take note of my other comment regarding System Pins above: https://github.com/ni/semi-test-library-dotnet/pull/574/changes#r3756405140

For this example, I would want this to look something more like this (note it's not necessary to demonstrate the single site number accepted constructor as you are trying to, it's not that meanful for this particular example):
```cs
// Dictionary containing site-unique data values, including a site-agnostic entry (-1).
// Note there can only be one site-agnostic value represented in a SiteData object.
var perSiteDataDictionary = new Dictionary<int, double> { [1] = 11, [2] = 22, [-1] = 33 };

// Create a new SiteData object containing the sites defined by the Keys of the dictionary. 
var siteData = new SiteData<double>(siteNumbers: perSiteDataDictionary.Keys);
// Add each dictionary item, including the site-agnostic entry represented by site -1.
foreach (var item in perSiteDataDictionary)
{
   siteData.SetValue(value: item.Value, siteNumber: item.Key);
}
```

Mattjet27 (changes requested):

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: Updated the code.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: For `SiteData`, we don't have a constructor that accepts multiple site numbers as of now. This is because of the ambiguity with another constructor that accepts `T[]`.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: Updated the name of the variable.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: Removed the comments.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
gmrsekar: I've updated the example to use the empty constructor since we don't have a constructor that takes multiple site numbers as of now.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: As of now, we only have constructors that take single or multiple pin names along with site numbers. We don't have a constructor that just takes pin names argument alone without any site numbers. 

Updated the example to use empty constructor and then add sites and pins using a single `AddSite` method.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the code.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the example. Other examples have specific system pin inputs, so they're taken care by default.

Mattjet27: @gmrsekar, when I went to test this, I realize it was compiling, but that's only because it was using a different overload and treating -1 as a value to set for all pins and sites. Therefore, we need to update this to the following:

```cs
// Per-pin, per-site data array.
// Each element in the perPinData array corresponds to a pin in the pinNames array.
var perPinData = new double[] { 1.5, 2.5, -22.5 };

// Note that data associated with system pins is considered site-agnostic,
// and site-agnostic data can represented with -1 as the site value.
// Typically, the site numbers array used within a code module will not explicitly contain a -1 value.
// Therefore, -1 will need to be appended to the siteNumbers array when creating a PinSiteData object that contains system pin data.
// This can be done using array operations or by converting the array into a List, but in most instances it is best not to modify the original siteNumbers array,
// as it may be used elsewhere in the code module where the system pin data is not relevant and can result in unintended consequences.
// The most efficient way to handle this is by first constructing the PinSiteData object with the known Pin Names and Site Numbers array,
// and then adding the -1 site manually with the AddSite method.
// Alternatively, the site and pins can simply be added dynamically when SetValue is called.
var pinSiteData = new PinSiteData<double>(pinNames: pinNames, siteNumbers: siteNumbers);
pinSiteData.AddSite(-1);
for (int i = 0; i < perPinData.Length; i++)
```

Examples/source/CodeSnippets/DataAbstraction/SiteDataExamples.cs
Mattjet27: @gmrsekar, that's right, disregard this comment.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
Mattjet27: > As of now, we only have constructors that take single or multiple pin names along with site numbers. We don't have a constructor that just takes pin names argument alone without any site numbers.

@gmrsekar, we should not need a separate constructor, the `public PinSiteData(string[] pinNames, params int[] siteNumbers)` constructor should allow the user to only pass in the pin names, as its method documentation states, "If the `siteNumbers` parameter is not provided, an empty `SiteData` object is associated with the pin names".

You should be able to implement it as follows:
```cs
// Use the pin-only constructor, then add the sites.
// This constructor allows you to declare pins upfront and add sites later on.
var pinSiteData = new PinSiteData<double>(pinNames: "VDET", "VCC1");
// Add sites
pinSiteData.AddSite(2, 4, 3);
```

Mattjet27 (changes requested):

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Ahh yes. The error I got was for a different reason. It expects argument to be in a string array format. So updated the example to pass a string array.

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Made some modifications to this code snippet to add -1 only to system pin and set values accordingly. Updated the example.

Mattjet27 (approved):

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
```diff
@@ -507,25 +507,30 @@ internal static void ConstructWithPinAndSiteUniqueDataArray()
 
         internal static void BuildWithPinAndSiteUniqueDataArray()
         {
+            // Site numbers to associate with the data.
+            var siteNumbers = new int[] { 2, 4, 3 };
+            // Pin names to associate with the data.
+            var pinNames = new string[] { "VDET", "VCC1" };
             // 2D jagged array of pin and site unique data,
             // where the first dimension represents pins (2) and the second dimension represents sites (3).
+            // Note that order of the elements in the perPinPerSiteData array must match the order of the pinNames and siteNumbers arrays. Given the out of order siteNumbers array, the data for site 3 is in the last column of the perPinPerSiteData array.
```
Mattjet27: @gmrsekar, instead of  "Given the out of order siteNumbers array,  ...", remove this and add a comment in line with each pin's data literal:
```cs
var perPinPerSiteData = new double[][]
{
    new double[] { 42, 105, 206 }, // VDET data for sites: 2, 4, 3.
    new double[] { 55, 2048, 0.5 } // VCC1 data for sites: 2, 4, 3.
};
```

Mattjet27 (changes requested):

Examples/source/CodeSnippets/DataAbstraction/PinSiteDataExamples.cs
gmrsekar: Updated the example.

Mattjet27 (approved):

PR #575, opened 2026-07-29
Adding Configure Output Extension Methods

hcnagaraj-e:

- Added 3 Configure Output Extension methods along with overloads for `SiteData` and `PinSiteData`
- Added 4 Unit Tests for each Output Extension method covering overloads.
- Methods are 
  - ConfigureOutputEnabled
  - ConfigureOutputImpedance
  - ConfigureOutputMode (No overloads)

FGen extension methods are added as part of usability improvements.

Feature:[ [STL] [New Instrument Support] [FGen] Support Standard Waveform Generation](https://dev.azure.com/ni/DevCentral/_workitems/edit/2631247)

Feature Spec:  [FGen Standard Waveform Generation Support](https://dev.azure.com/ni/DevCentral/_wiki/wikis/AppCentral.wiki/165216/FGen-Standard-Waveform-Generation-Support)

Work Item: [Implement Configure Output Extensions for FGen](https://dev.azure.com/ni/DevCentral/_workitems/edit/3861474)

- [x] Tested running unit tests on Offline mode as well as with real hardware
- [x] Ensured general guidelines are followed
- [x] Tested for both single pin and multiple pins cases


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <param name="sessionsBundle">The <see cref="FgenSessionsBundle"/> object.</param>
+        /// <param name="outputMode">Specifies the output mode that you want the signal generator to use.</param>
+        /// <remarks>
+        /// As of STL 26.5, only the `Function` output mode is supported. Attempting to configure any other output mode throws an exception.
+        /// </remarks>
+        public static void ConfigureOutputMode(this FgenSessionsBundle sessionsBundle, OutputMode outputMode)
+        {
+            if (outputMode != OutputMode.Function)
+            {
+                throw new NISemiconductorTestException($"Output mode {outputMode} is not supported.");
```
gmrsekar: I think we need to use resource strings to add exception messages and call the resource string while throwing exception.

TestAssets/Supporting Materials/Pin Maps/FgenSingleInstrumentPerPin.pinmap
```diff
@@ -0,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<PinMap schemaVersion="1.1" xmlns="http://www.ni.com/TestStand/SemiconductorModule/PinMap.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<Instruments>
```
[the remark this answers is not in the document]
hcnagaraj-e: It's created from Pin Map Editor, this is the correct format

TestAssets/Supporting Materials/Pin Maps/FgenSingleInstrumentPerSite.pinmap
```diff
@@ -0,0 +1,39 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<PinMap schemaVersion="1.1" xmlns="http://www.ni.com/TestStand/SemiconductorModule/PinMap.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<Instruments>
```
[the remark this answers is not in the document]
hcnagaraj-e: It's created from Pin Map Editor, this is the correct format

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformConfigureOutputModeOperation_Succeeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.ConfigureOutputMode(OutputMode.Function);
+        }
```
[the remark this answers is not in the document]
hcnagaraj-e: Added assertion part

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: I have used resource strings now, no hardcoded string data

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Linq;
```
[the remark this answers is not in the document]
hcnagaraj-e: Removed

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,280 @@
...
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            foreach (var outputMode in Enum.GetValues(typeof(OutputMode)))
+            {
+                if ((OutputMode)outputMode != OutputMode.Function)
+                {
+                    var exception = Assert.Throws<NISemiconductorTestException>(() => sessionsBundle.ConfigureOutputMode((OutputMode)outputMode));
+                    Assert.Contains(string.Format(CultureInfo.InvariantCulture, ResourceStrings.FGen_InvalidOutputModeException, outputMode), exception.Message);
```
nitamil, line 229: Can you split the Assert in another loop.

nitamil (changes requested):
Please make the test names consistent with AAA  model.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: Anish, we are not setting anything here. If we need to assert outside this loop, we need to catch every exception and build into an array.  We need to catch and assert immediately

hcnagaraj-e: Please feel free to re-open this comment if you have specific feedback. Based on Ashutosh's suggestion, I made it AA format as we cannot separate assert and action

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <remarks>
+        /// This method specifies the output impedance of the NI signal generator at the output connector.
+        /// NI signal generators have an output impedance of 50 Ω and an optional 75 Ω on select modules.
+        /// <para>
+        /// If the load impedance value matches the output impedance, the voltage at the signal output connector is at the necessary level.
+        /// The voltage at the signal output connector varies with load output impedance, up to doubling the voltage for a high-impedance load.
+        /// </para>
+        /// <para>
+        /// You cannot change terminal configuration while the device is generating a waveform.
+        /// If you want to change the device configuration, call <see cref="Control.Abort" crefType= "undefined"/> or wait for the generation to complete.
```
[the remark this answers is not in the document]
hcnagaraj-e: It should work once the other PR is merged. Keeping this comment open.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,301 @@
...
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerSite.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        [InlineData("FgenSingleInstrumentSharedAcrossPinsAndSites.pinmap")]
+        public void InitializeBundleWithMultiplePins_PerformConfigureOutputEnabledOperationWithSiteData_Succeeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var pinNames = new string[] { "A", "B" };
+            var sessionsBundle = sessionManager.Fgen(pinNames);
+
```
gmrsekar: Can we delete this empty line as lines from 74 to 80 are from Arrange part?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <para>
+        /// If the load impedance value matches the output impedance, the voltage at the signal output connector is at the necessary level.
+        /// The voltage at the signal output connector varies with load output impedance, up to doubling the voltage for a high-impedance load.
+        /// </para>
+        /// <para>
+        /// You cannot change terminal configuration while the device is generating a waveform.
+        /// If you want to change the device configuration, call 'Abort' extension method or wait for the generation to complete.
+        /// </para>
+        /// </remarks>
+        public static void ConfigureOutputImpedance(this FgenSessionsBundle sessionsBundle, double impedance = 50)
```
nitamil, line 65: if there are only 2 values 50 ohms and 75 ohms. Can we define an enum for those values.

nitamil (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: The FGen device used STS only supports 50 ohms. I have made default value as 50 ohms. Since, driver is exposing this property as double, I am retaining the same signature. 
Even in Feature Spec, we called out the method signature which uses Double datatype.

For now, I want to proceed with the same signature that is called out in feature spec. We can discuss with Matt and modify it later if needed.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        /// <inheritdoc cref="ConfigureOutputEnabled(FgenSessionsBundle, bool)"/>
+        public static void ConfigureOutputEnabled(this FgenSessionsBundle sessionsBundle, PinSiteData<bool> outputEnable)
+        {
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            {
+                sessionInfo.Session.Output.SetEnabled(sitePinInfo.IndividualChannelString.Split('/').Last(), outputEnable.GetValue(sitePinInfo));
+            });
+        }
```
Ashutosh-NI, line 46: I think each of `ConfigureOutputEnabled` should override the `<param>` for `outputEnabled` as each of them take different "type" of `outputEnabled`

```csharp
 /// <inheritdoc cref="ConfigureOutputEnabled(FgenSessionsBundle, bool)"/>
 /// <param name="sessionsBundle"/>
/// <param name="outputEnable">taking about the type of parameter.</param>
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+                sessionInfo.Session.Output.SetEnabled(sessionInfo.AllChannelsString, outputEnable);
+            });
+        }
+
+        /// <inheritdoc cref="ConfigureOutputEnabled(FgenSessionsBundle, bool)"/>
+        public static void ConfigureOutputEnabled(this FgenSessionsBundle sessionsBundle, SiteData<bool> outputEnable)
+        {
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            {
+                sessionInfo.Session.Output.SetEnabled(sitePinInfo.IndividualChannelString.Split('/').Last(), outputEnable.GetValue(sitePinInfo.SiteNumber));
```
Ashutosh-NI, line 35: Just want to know why are we splitting the individual string?
`sitePinInfo.IndividualChannelString.Split('/').Last()`

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
```diff
@@ -248,5 +248,14 @@ internal static string Digital_TMUUnsupportedPolarity {
                 return ResourceManager.GetString("Digital_TMUUnsupportedPolarity", resourceCulture);
             }
         }
```
Ashutosh-NI: Are these changes intentional?

SemiconductorTestLibrary.Extensions/source/ResourceStrings.resx
```diff
@@ -180,4 +180,7 @@
   <data name="Digital_TMUUnsupportedPolarity" xml:space="preserve">
     <value>Unsupported Polarity.</value>
   </data>
-</root>
+  <data name="FGen_InvalidOutputModeException" xml:space="preserve">
+    <value>Output mode {0} is not supported, only 'Function' is supported in STL.</value>
```
Ashutosh-NI: Should we put single quotes or quotes around the input function type?

```suggestion
    <value>Output mode '{0}' is not supported, only 'Function' is supported in STL.</value>
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
Ashutosh-NI, line 65: Can we add the validation for `impedance` type early in the function. As we are sure that only 50 is supported `impedance` , we can validate that before giving it to driver as that will be much more performant.

Ashutosh-NI (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: Unlike DC Power, FGen does not support InstrumentName/ChannelName format. We need to give channel name only which will be '0' or '1'

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: @Ashutosh-NI  Supported `impedance` can change based on instrument model. I feel we should not restrict it from STL side. If we are going in that direction, there is no need to provide any extension method for setting impedance.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: @Ashutosh-NI Even if it takes different format, it's still the same input, right? do we really need to override <param> ?

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
hcnagaraj-e: @Ashutosh-NI I am not sure what you are referring, this file is autogenerated

SemiconductorTestLibrary.Extensions/source/ResourceStrings.resx
hcnagaraj-e: Added single quotes

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: I referred DC power implementation; parameter description is same for other 2 overloads

image

https://github.com/ni/semi-test-library-dotnet/blob/main/SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs

hcnagaraj-e:
@copilot resolve the merge conflicts in this pull request

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
Ashutosh-NI, line 1: @hcnagaraj-e Can we add hardware configuration to either individual tests (as many as possible) or whole class (if all the tests can be run in hardware).
Tags:
```csharp
        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.GP3))]
        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.Lungyuan))]
        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <summary>
+        /// Configures the output impedance of the signal generator.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="FgenSessionsBundle"/> object.</param>
+        /// <param name="impedance">Specifies the impedance value that you want the signal generator to use.</param>
+        /// <remarks>
+        /// This method specifies the output impedance of the NI signal generator at the output connector.
+        /// NI signal generators have an output impedance of 50 Ω and an optional 75 Ω on select modules.
+        /// <para>
+        /// If the load impedance value matches the output impedance, the voltage at the signal output connector is at the necessary level.
```
Ashutosh-NI, line 57: can you add line bread `<br/>` at end of the lines which you want to be in next lines for readibilty in intelisense.

```suggestion
        /// If the load impedance value matches the output impedance, the voltage at the signal output connector is at the necessary level.<br/>
```

Ashutosh-NI: And this can be done in all the other required places.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
```diff
@@ -10,16 +10,16 @@
 
 namespace NationalInstruments.SemiconductorTestLibrary {
     using System;
-
-
+    
+
```
Ashutosh-NI, line 14: Is this line/space intentional ?

Image

Ashutosh-NI: Similarly, in 

Image

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
```diff
@@ -338,15 +347,13 @@ internal static string Digital_TMUUnsupportedPulseWidth {
                 return ResourceManager.GetString("Digital_TMUUnsupportedPulseWidth", resourceCulture);
             }
         }
-
+        
         /// <summary>
-        ///   Looks up a localized string similar to The publishDataIdFormatter parameter cannot be null or empty and must contain a format placeholder &apos;{{0}}&apos;..
+        ///   Looks up a localized string similar to Output mode &apos;{0}&apos; is not supported, only &apos;Function&apos; is supported in STL..
```
Ashutosh-NI: We generally do not use "STL" as abbreviation in exception/doumentaion. If not required can we remove "STL"?

```suggestion
        ///   Looks up a localized string similar to Output mode &apos;{0}&apos; is not supported, only &apos;Function&apos; is supported..
```

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System.Globalization;
+using System.Linq;
+using NationalInstruments.ModularInstruments.NIFgen;
+using NationalInstruments.SemiconductorTestLibrary.Common;
+using NationalInstruments.SemiconductorTestLibrary.DataAbstraction;
+
+namespace NationalInstruments.SemiconductorTestLibrary.InstrumentAbstraction.Fgen
+{
+    /// <summary>
+    ///  Defines methods for output configurations.
```
Ashutosh-NI: can we add fgen in the summary?

```suggestion
    ///  Defines methods for NI-Fgen output configurations.
```

Ashutosh-NI (approved):
Approved with suggestions.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: @Ashutosh-NI I have created Paragraphs. This particular paragraph meant to be a single line text.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
hcnagaraj-e: I did not do any changes manually, autogenerated file has these spaces. I will try to recreate again and check

hcnagaraj-e: The spaces added to till the indentation level. This is how it is present for all line gaps. Looks like the one which are shown in diff are the result of manually removing those spaces in previous updates.

Even the positions of the functions are now adjusted based on alphabetical order. 

So, diff is correct. I also verified it by regenerating it again.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
hcnagaraj-e: @Ashutosh-NI I have removed the STL part as suggested

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: I have added [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]

hcnagaraj-e: I have added STSNIB Cauvery trait

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <param name="sessionsBundle">The <see cref="FgenSessionsBundle"/> object.</param>
+        /// <param name="outputEnable">Specifies the state of the output enable relay. Set outputEnable to <see langword="true"/> to enable the relay.</param>
+        /// <remarks>
+        /// Use this method to configure whether the signal that the signal generator produces appears at the output connector.
+        /// </remarks>
+        public static void ConfigureOutputEnabled(this FgenSessionsBundle sessionsBundle, bool outputEnable)
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.Output.SetEnabled(sessionInfo.AllChannelsString, outputEnable);
```
NikhilSivadas-NIB, line 26: API reference state there could be additional OperationNotSupportedException and SelectorNameException that could occur for SetEnabled, can you have a look and document STL exception if its necessary ?

Image

NikhilSivadas-NIB, line 26: Also, can we include unit test for these exceptions ?

NikhilSivadas-NIB, line 26: is there a specific reason why we use AllChannelsString string here? example code use IndividualChannelString

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+            });
+        }
+
+        /// <inheritdoc cref="ConfigureOutputEnabled(FgenSessionsBundle, bool)"/>
+        public static void ConfigureOutputEnabled(this FgenSessionsBundle sessionsBundle, SiteData<bool> outputEnable)
+        {
+            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            {
+                sessionInfo.Session.Output.SetEnabled(sitePinInfo.IndividualChannelString.Split('/').Last(), outputEnable.GetValue(sitePinInfo.SiteNumber));
```
NikhilSivadas-NIB, line 35: Seems sitePinInfo.IndividualChannelString.Split('/').Last() repeated several times. If this logic is expected to be reused across FGen APIs, can we move it to a shared internal helper/extension method? Otherwise, a private helper within this file could improve readability and centralize the parsing logic.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <para>
+        /// You cannot change terminal configuration while the device is generating a waveform.
+        /// If you want to change the device configuration, call 'Abort' extension method or wait for the generation to complete.
+        /// </para>
+        /// </remarks>
+        public static void ConfigureOutputImpedance(this FgenSessionsBundle sessionsBundle, double impedance = 50)
+        {
+            sessionsBundle.Do(sessionInfo =>
+            {
+                sessionInfo.Session.Output.SetImpedance(sessionInfo.AllChannelsString, impedance);
```
NikhilSivadas-NIB, line 69: Similar to previous, can we think of need to document this if necessary ?

Image

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        [InlineData("FgenSingleInstrumentPerSite.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        [InlineData("FgenSingleInstrumentSharedAcrossPinsAndSites.pinmap")]
+        public void InitializeBundleWithMultiplePins_PerformConfigureOutputImpedanceOperationWithSiteData_Succeeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen(new string[] { "A", "B" });
+            var pinNames = new string[] { "A", "B" };
+            var siteNumbers = new int[] { 0, 1 };
+            var siteDataArray = new double[] { 50, 50 };
```
NikhilSivadas-NIB, line 155: can we use different values here to ensure different site-specific correctness?

NikhilSivadas-NIB (commented):
I have partially reviewed the PR, can you have a look at comments?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: Unfortunately, the FGen model we use seem to support only 50 acceptable values. I will cross check once again and confirm. Driver was throwing error when we try to set any other value.

We can check with Matt and see, if we still need this Extension method.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: @NikhilSivadas-NIB  I think we need to make a decision on whether we need to call out Driver Exceptions. Can you take up with Lorna and confirm. In the past, we only documented exception thrown by STL code. 

Even when writing unit tests, we wrote tests only for STL positive and negative test cases. Driver behavior can change over different versions. If we are testing driver functionality, we should put them under system tests. 

Also, channel names are abstracted from user code.  It does not make sense to document `SelectorNameException`. STL should take care of giving right channel name.

hcnagaraj-e: @NikhilSivadas-NIB In the past we did not write test for driver exceptions. It should not matter for STL whether driver throws exception are not. 
We cannot hit `SelectorNameException.` with test code using regular STL code/API. We need to call driver API deliberately with wrong channel names.

hcnagaraj-e: In the POC, we showed that how we can call driver APIs when channel name is accepted. But some of the APIs support providing channel names as comma separated values. In that case we can give AllChannelsString directly.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: I can make this as private helper method and move to a common UtilityHelper class.
We have other options
1. Store channel names as individual channel string instead of storing instrument name/channel name. But this behavior can change in 27.5 when we update FGen driver.
2. Add additional filed to `SitePinInfo` class to store just the channel name.
3. Extend SitePinInfo and add GetChannelName extension method.

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
```diff
@@ -10,16 +10,16 @@
...
+    
     /// <summary>
     ///   A strongly-typed resource class, for looking up localized strings, etc.
     /// </summary>
     // This class was auto-generated by the StronglyTypedResourceBuilder
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "18.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
```
NikhilSivadas-NIB, line 22: is this intended change?

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
```diff
@@ -59,7 +59,7 @@ internal ResourceStrings() {
                 resourceCulture = value;
             }
         }
-
+
```
NikhilSivadas-NIB, line 62: look some new line added here and below, is that intentional ?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        {
+            InitializeAndClose.Close(_tsmContext);
+        }
+
+        #region ConfigureOutputEnabled tests
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformConfigureOutputEnabledOperation_Succeeds(string pinmap)
```
NikhilSivadas-NIB, line 40: Can we rename all tests? Act part to start with public api method as mentioned in previous PR

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
NikhilSivadas-NIB: Here impedance value we are letting to pass without any validation so it would throw exception which would  wrapped in STLException

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        [Theory]
+        [InlineData("FgenSingleInstrumentPerPin.pinmap")]
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        public void InitializeBundleWithSinglePin_PerformConfigureOutputImpedanceOperation_Succeeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen("A");
+
+            sessionsBundle.ConfigureOutputImpedance(50);
```
NikhilSivadas-NIB, line 120: Looks all the test cases are positive cases only, can we add negative and exception cases covered in unit tests?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        [InlineData("FgenSingleInstrumentPerPinPerSite.pinmap")]
+        [InlineData("FgenSingleInstrumentSharedAcrossPinsAndSites.pinmap")]
+        public void InitializeBundleWithMultiplePins_PerformConfigureOutputImpedanceOperation_Succeeds(string pinmap)
+        {
+            var sessionManager = Initialize(pinmap);
+            var sessionsBundle = sessionManager.Fgen(new string[] { "A", "B" });
+
+            sessionsBundle.ConfigureOutputImpedance(50);
+
+            AssertOutputImpedance(sessionsBundle, 50);
```
NikhilSivadas-NIB, line 139: As mentioned can we add coverage for non default and wrong value?

NikhilSivadas-NIB (commented):
Added few more comments, can you have a look?

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
hcnagaraj-e: This is auto generated file. It's not new empty line; it is adding spaces as per the indentation. All line gaps have spaces. Somebody has removed these autogenerated spaces in previous PR. Most of the other line gaps have these spaces. This is fixing those things,

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: @NikhilSivadas-NIB  -Ve exception cases are for driver and Not for STL. I think we need to make a decision on whether we need to cover driver testing or just limit it to STL testing.

The unsupported impedance is purely based on Model of the FGen, we can assume model as PXI-5433 and write -ve test case but that is not right. If we are doing that then we need to add new folder called system tests and move those tests into that folder.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: @NikhilSivadas-NIB Looks like for PXI-5433 model, any value other than 50 ohms is a wrong value. Can you check with Matt on this

SemiconductorTestLibrary.Extensions/source/ResourceStrings.Designer.cs
hcnagaraj-e: This is autogenerated. I took a clean file from Main and add new entry; it is updating the version to 17.0.0.0. Looks like somebody edited this file manually in previous PRs. I verified this multiple time, it needs to be reset to 17.0.0.0

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: @NikhilSivadas-NIB I have responded this in previous comment, Can we check with Matt and confirm if we need to document driver exception?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: @NikhilSivadas-NIB The FGen model supported in STS is PXI-5433 2-Channel. This model only supports 50 impedance value.
@Mattjet27 Since only 50 is supported, should we keep this extension method or, should we consider removing it altogether?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: I am closing this comment, will update the exception tag for the methods which likely result in exceptions

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: I have added negative test based on driver behavior. Setting unsupported impedance value.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: I have added exception tag for `ConfigureImpedance`. That is the only method which can lead to driver exception in a regular use case.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/FGen/Output.cs
hcnagaraj-e: Added CommonUtils class and added helper method.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/FGen/OutputTests.cs
hcnagaraj-e: Renamed all tests with public API name.

PR #576, opened 2026-07-30
DC Power Extension Methods | Source | Measure Methods

Shalini-Subramanian:
This pull request adds comprehensive support for measuring and querying compliance (incompliance) status alongside voltage and current measurements in the DCPower instrument abstraction. It introduces new API methods to obtain both the measurement and its compliance result per pin and per site, updates the measurement logic to handle both OnDemand and OnMeasureTrigger modes, and significantly expands unit test coverage for these new capabilities. Additionally, a new pin map file is added for more thorough test scenarios.

**API Enhancements:**

* Added new extension methods to `DCPowerSessionsBundle` for measuring current and voltage with compliance status (`MeasureCurrentWithInCompliance`, `MeasureVoltageWithInCompliance`) and for querying compliance status alone (`QueryInCompliance`). These return results as pin- and site-aware data structures, pairing each measurement with its compliance result.

* Implemented a private helper method `MeasureWithInCompliance` to generalize the logic for pairing measurements with compliance results, supporting both voltage and current selection.

**Measurement Logic Improvements:**

* Introduced `MeasureVoltageAndCurrentWithInCompliance` in `DCPowerSessionInformation`, which efficiently collects voltage, current, and compliance results for both OnDemand and OnMeasureTrigger configurations, using optimized driver calls and correct result mapping.

* Added extensive unit tests to cover all new compliance-aware measurement and query methods with OnDemand and Non-OnDemand trigger modes.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -600,6 +642,59 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
...
+            if (onDemandChannels.Any())
+            {
+                var onDemandChannelsString = string.Join(",", onDemandChannels.Select(c => c.Item1.IndividualChannelString));
+                var measureResult = session.Measurement.Measure(onDemandChannelsString);
+                for (int i = 0; i < onDemandChannels.Count; i++)
+                {
+                    int index = onDemandChannels[i].Item2;
+                    voltageMeasurements[index] = measureResult.VoltageMeasurements[i];
+                    currentMeasurements[index] = measureResult.CurrentMeasurements[i];
+                    inComplianceResults[index] = session.Measurement.QueryInCompliance(onDemandChannels[i].Item1.IndividualChannelString);
```
nitamil, line 683: Does the QueryInCompliance has to be per channel? does it not accept the CSV list of channels? if it accepts please move the function call outside the for loop similar to `Measure`. Performance is the key factor here. so instead of multiple driver calls it will be fast and efficient if we have just one driver call.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -600,6 +642,59 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
...
+        /// <summary>
+        /// Measures the voltage and current, and queries the incompliance result.
+        /// </summary>
+        /// <param name="sessionInfo">The <see cref="DCPowerSessionInformation"/> object.</param>
+        /// <returns>The measurements. Item1 is voltage measurements, Item2 is current measurements, Item3 is incompliance results.</returns>
+        /// <remarks>
+        /// For channels configured with OnMeasureTrigger, the incompliance result is obtained via Fetch.
+        /// For channels configured with OnDemand, the incompliance result is queried via QueryInCompliance.
+        /// </remarks>
+        public static Tuple<double[], double[], bool[]> MeasureVoltageAndCurrentWithInCompliance(this DCPowerSessionInformation sessionInfo)
```
nitamil, line 654: Can we rename it to "MeasureVoltageCurrentAndInCompliance"?

@Mattjet27 @lornazh @NikhilSivadas-NIB , This is a public API so need your thoughts and suggesions for this function.

nitamil (changes requested):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
nitamil: @Shalini-Subramanian For now rename this function as `MeasureVoltageCurrentAndInCompliance`.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: `Measure(channelString)` accepts a CSV string (e.g., "ch0,ch1,ch2") and returns arrays of measurements
`QueryInCompliance(channelString)` accepts only a single channel string and returns a single boolean value (not an array)

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: Updated

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -234,6 +234,48 @@ public static PinSiteData<double> MeasureCurrent(this DCPowerSessionsBundle sess
...
+        }
+
+        /// <summary>
+        /// Queries whether the target pin(s) are currently in compliance and returns a pin- and site-aware data object.
+        /// </summary>
+        /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
+        /// <returns>The per-pin per-site compliance status.</returns>
+        public static PinSiteData<bool> QueryInCompliance(this DCPowerSessionsBundle sessionsBundle)
+        {
+            sessionsBundle.ClearBacklogIfSoftwareEdgeTrigger();
```
ni-ramya, line 385: Is the required in query compliance? A query should do only a query but here it clears and then queries. Is this needed? If clear is needed, it should be handled by the user, IMO... If this should be a part of this API, this behaviour should atleast be documented appropriately in the xml documentation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -600,6 +642,59 @@ public static Tuple<double[], double[]> MeasureVoltageAndCurrent(this DCPowerSes
...
+        /// <summary>
+        /// Measures the voltage and current, and queries the incompliance result.
+        /// </summary>
+        /// <param name="sessionInfo">The <see cref="DCPowerSessionInformation"/> object.</param>
+        /// <returns>The measurements. Item1 is voltage measurements, Item2 is current measurements, Item3 is incompliance results.</returns>
+        /// <remarks>
+        /// For channels configured with OnMeasureTrigger, the incompliance result is obtained via Fetch.
+        /// For channels configured with OnDemand, the incompliance result is queried via QueryInCompliance.
+        /// </remarks>
+        public static Tuple<double[], double[], bool[]> MeasureVoltageCurrentAndInCompliance(this DCPowerSessionInformation sessionInfo)
```
ni-ramya, line 847: This method does exactly what `MeasureVoltageAndCurrent` does except the compliance query. This leads to duplication and maintenance overhead. Can we update `MeasureVoltageAndCurrent` to use `MeasureVoltageAndCurrentWithInCompliance` internally, but discard compliance results?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -1147,6 +1147,147 @@ void MeasureTest()
...
+        [InlineData(true)]
+        [InlineData(false)]
+        public void DifferentSMUDevicesOneChannelMeasureOnTriggerOthersMeasureOnDemand_ForceVoltageMeasureVoltageWithInCompliance_ReturnsCorrectValue(bool pinMapWithChannelGroup)
+        {
+            var sessionManager = Initialize(pinMapWithChannelGroup);
+            var sessionsBundle = sessionManager.DCPower("VDD");
+            var voltageLevel = 3.6;
+            string firstChannelString = sessionsBundle.InstrumentSessions.ElementAt(0).AssociatedSitePinList[0].IndividualChannelString;
+            sessionsBundle.ConfigureMeasureWhen(DCPowerMeasurementWhen.OnDemand);
+            sessionsBundle.InstrumentSessions.ElementAt(0).Session.Outputs[firstChannelString].Measurement.MeasureWhen = DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete;
```
ni-ramya, line 1858: Should this be `OnMeasureTrigger` or the test name should be updated to 'DifferentSMUDevicesOneChannelAutoAfterSourceCompleteOhers....'

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -1229,5 +1370,56 @@ private void AssertResultAssociatedWithPinGroupName(PinSiteData<double> results,
...
         }
+
+        private void AssertAllChannelsHaveCorrectResult(PinSiteData<Tuple<double, bool>> results, double expectedMeasurement)
+        {
+            foreach (var siteNumber in results.SiteNumbers)
+            {
+                foreach (var pin in results.PinNames)
+                {
+                    var value = results.GetValue(siteNumber, pin);
+                    Assert.Equal(expectedMeasurement, value.Item1, precision: 3);
```
ni-ramya: precision 3 will be insufficient for current measurements whihc are in uA. Have precision 6.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -1147,6 +1147,147 @@ void MeasureTest()
             dcPower.UngangPinGroup("MergedPowerPins");
         }
 
+        [Theory]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        [InlineData(DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete, true)]
+        [InlineData(DCPowerMeasurementWhen.OnDemand, false)]
```
ni-ramya, line 1833: Should we alos test for 'OnMeasureTrigger'?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -1147,6 +1147,147 @@ void MeasureTest()
...
+
+            var results = sessionsBundle.MeasureVoltageWithInCompliance();
+
+            AssertAllChannelsHaveCorrectResult(results, voltageLevel);
+        }
+
+        [Theory]
+        [Trait(nameof(HardwareConfiguration), nameof(HardwareConfiguration.STSNIBCauvery))]
+        [InlineData(DCPowerMeasurementWhen.AutomaticallyAfterSourceComplete, true)]
+        [InlineData(DCPowerMeasurementWhen.OnDemand, false)]
```
ni-ramya, line 1887: Should we alos test for 'OnMeasureTrigger'?

ni-ramya (approved):

ni-ramya (changes requested):
I would like to review the changes requested.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: But performance wise, I would say querying the incompliance would be not optimal. And if we are replacing the MeasureVoltageAndCurrent then we may need to do stress and performance testing

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: @nitamil Can you please let me know your thoughts on this

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: Shall I introduce a new optional parameter `bool queryIncompliance = false` to the method `MeasureVoltageAndCurrent` and fill in `inComplianceResults[index]` 

@nitamil please let me know if this fine

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
Shalini-Subramanian: No,

In this test we are configure the measure differently for 2 different channels.

1. AutomaticallyAfterSourceComplete for 1st channel
2. OnDemand

Here `AutomaticallyAfterSourceComplete` falls under the category `MeasureOnTrigger` [nonDemand] where as `OnDemand` falls under `MeasureOnDemand`.

Thus renaming is optional

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
Shalini-Subramanian: `OnMeasureTrigger` comes under non-demand category and falls under the test `DifferentSMUDevicesOneChannelMeasureOnTriggerOthersMeasureOnDemand_ForceCurrentMeasureCurrentWithInCompliance_ReturnsCorrectValue`

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
nitamil: Please create a NI Stack Overflow discussion for the same and confirm it is expected behaviour or a bug. Based on that we can file a bug.  Also confirm if the behaviour is matching the documentation or not.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
nitamil: This is fine because we are clearing for specific measurement case.  To be precise before sending the software trigger we are clearing previous records only for software triggered measurement.  As the software trigger will be sent during the measurement operation.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
nitamil: What will be the return parameter in that case? Also it might break the forward/backward compatibility.  Even adding a overload method  will have some design challenges i think.  I am currently aligned with the current Public API as in this PR. we can modify and improve the internal implementation to reduce the repeated code logic.

Feel free to share your ideas over a POC code and get the approval from @Mattjet27  or @lornazh  or @NikhilSivadas-NIB.  That will address some of the open questions that we have.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
nitamil: If we have ganging test code, that will cover the OnMeasureTrigger test cases. As ganging uses the `OnMeasureTrigger` settings for the follower channels.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: Sure Anish, Thanks for the clarification. Given that, I am resolving this comment and thus not removing the clear

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian:
```cs
public static Tuple<double[], double[], bool[]> MeasureVoltageCurrentAndIncompliance(this DCPowerSessionInformation sessionInfo, bool queryIncompliance = false)
{
    var session = sessionInfo.Session;
    List<SitePinInfo> listOfChannelsToMeasure = sessionInfo.AssociatedSitePinList.Where(sitePin => !sitePin.SkipOperations).ToList();
    SplitChannels(session, listOfChannelsToMeasure, out var onDemandChannels, out var nonOnDemandChannels);
    int channelCount = listOfChannelsToMeasure.Count;
    var voltageMeasurements = new double[channelCount];
    var currentMeasurements = new double[channelCount];
    var inComplianceResults = new bool[channelCount];

    foreach (var channel in nonOnDemandChannels)
    {
        if (channel.Item3.Measurement.MeasureWhen == DCPowerMeasurementWhen.OnMeasureTrigger
            && channel.Item3.Triggers.MeasureTrigger.Type == DCPowerMeasureTriggerType.SoftwareEdge)
        {
            channel.Item3.Triggers.MeasureTrigger.SendSoftwareEdgeTrigger();
        }
    }

    // Measure all on demand channels in a single driver call to optimize test time.
    if (onDemandChannels.Any())
    {
        var onDemandChannelsString = string.Join(",", onDemandChannels.Select(c => c.Item1.IndividualChannelString));
        var measureResult = session.Measurement.Measure(string.Join(",", onDemandChannelsString));
        for (int i = 0; i < onDemandChannels.Count; i++)
        {
            int index = onDemandChannels[i].Item2;
            voltageMeasurements[index] = measureResult.VoltageMeasurements[i];
            currentMeasurements[index] = measureResult.CurrentMeasurements[i];
            if (queryIncompliance)
            {
                inComplianceResults[index] = session.Measurement.QueryInCompliance(onDemandChannels[i].Item1.IndividualChannelString);
            }
        }
    }

    foreach (var channel in nonOnDemandChannels)
    {
        var fetchResult = session.Measurement.Fetch(channel.Item1.IndividualChannelString, new PrecisionTimeSpan(20), 1);
        voltageMeasurements[channel.Item2] = fetchResult.VoltageMeasurements[0];
        currentMeasurements[channel.Item2] = fetchResult.CurrentMeasurements[0];
        if (queryIncompliance)
        {
            inComplianceResults[channel.Item2] = fetchResult.InCompliance[0];
        }
    }

    return new Tuple<double[], double[], bool[]>(voltageMeasurements, currentMeasurements, inComplianceResults);
}
```

This is how the existing method would look like. 

I agree to the point `it might break the forward/backward compatibility. Even adding a overload method will have some design challenges` and thinking if we can step back on the change suggested by Ramya and have the API in this current PR. Please let me know your thoughts

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: @nitamil Based on the documentation, it is expected that `QueryIncompliance` driver call returns whether the specified channel is in compliance. And it does not return multiple values.

Reference: [NI.doc ref](https://www.ni.com/docs/en-US/bundle/ni-dcpower-csharp-api-ref/page/nationalinstruments-modularinstruments-nidcpower-dcpowermeasurement-queryincompliance__string.html#d40871e230)

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: This would end up with updating the already existing public API's, which is not a good practice. Thus closing this thread.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: Closing this thread, as querying multiple channels is not supported in the driver rn

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
Shalini-Subramanian: Added a new test case

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -278,6 +278,59 @@ public static PinSiteData<double> MeasureCurrent(this DCPowerSessionsBundle sess
...
+                return sessionsBundle.DoAndReturnPerSitePerPinResults((sessionInfo, sitePinInfo) =>
+                {
+                    return sessionInfo.Session.Measurement.QueryInCompliance(sitePinInfo.IndividualChannelString);
+                });
+            }
+
+            return sessionsBundle.DoAndReturnPerSitePerPinResults(
+                (sessionInfo, sitePinInfo) => sessionInfo.Session.Measurement.QueryInCompliance(sitePinInfo.IndividualChannelString),
+                caseDescription: string.Empty,
+                InCompliancePinSiteResultsFilling);
```
ni-ramya, line 397: This is not defined anywhere. Am I missing something?

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices;
```
ni-ramya: Lets remove this if unused

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
Shalini-Subramanian: This piece of code is present in the Abstraction layer [PR Reference](https://dev.azure.com/ni/DevCentral/_git/srd-sebu-mixed-signal-apt/pullrequest/1373383).

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/DCPower/MeasureTests.cs
Shalini-Subramanian: Removed

ni-ramya (approved):

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Measure.cs
```diff
@@ -278,6 +278,59 @@ public static PinSiteData<double> MeasureCurrent(this DCPowerSessionsBundle sess
...
+        /// </summary>
+        /// <remarks>
+        /// For channels configured with OnMeasureTrigger, the results are obtained via Fetch, which includes the incompliance result.
+        /// For channels configured with OnDemand, the current is measured via Measure and the incompliance result is queried via QueryInCompliance.
+        /// </remarks>
+        /// <param name="sessionsBundle">The <see cref="DCPowerSessionsBundle"/> object.</param>
+        /// <param name="gangedPinsAsGroup">When false (default), each channel in a ganged/merged pin group is returned under its individual pin name. When true, the group is returned as a single value under the pin group name (current accumulated, in compliance only if all channels are in compliance).</param>
+        /// <returns>The per-pin per-site current measurements and incompliance results. Item1 is the current measurement, Item2 is the incompliance result.</returns>
+        public static PinSiteData<Tuple<double, bool>> MeasureCurrentWithInCompliance(this DCPowerSessionsBundle sessionsBundle, bool gangedPinsAsGroup = false)
```
TamilNeram: Please rename this function as MeasureCurrentAndInCompliance. The word `with` and `in` makes the definition bit confusing, like measure current only within compliance kind of meaning. To avoid confusion use `And`.

Shalini-Subramanian: Fixed

nitamil (approved):

NikhilSivadas-NIB (approved):
PR reviewed by @ni-ramya and @ni-ramya

PR #592, opened 2026-08-19
Update Abstraction dll from export 26.5.0d60

Utkarsh-NI-Emerson:
Please pull this in!


NikhilSivadas-NIB (approved):
Approved

Ashutosh-NI:
@NikhilSivadas-NIB We need to force merge this PR because it contains changes to the TmuContextManager class, including updates to method names.
The current intake validation failure is expected because the extensions project is still referencing the old Add method names from the version currently available in Git.
The required updates to those method references are included in PR #561. However, PR #561 depends on the abstractions introduced in this PR, creating a circular dependency between the two PRs.
To break this dependency and allow both changes to be integrated, we need to force merge this PR first.

CC: @Utkarsh-NI-Emerson @hcnagaraj-e 
Reference for the failure of the pr intake of this PR:
image

NikhilSivadas-NIB (approved):
Approved

PR #598, opened 2026-08-27
Ensure TMU cleanup on clear failure

Ashutosh-NI:
This pull request makes sure assigned TMU resources are always released when a TMU operation fails.

TMU assignment is virtual bookkeeping tracked in the static, process-lifetime `TMUContextManager`. Previously, if any step after `AssignTMUResources` threw, the exception propagated and the caller's `DisableTMU` / `ClearTMUAssignment` cleanup never ran. Those assignments were never released, so every later `AssignTMUResources` failed with `Digital_TMUNotEnoughResources` even though the hardware was free. On a device with only 2 TMUs, a single transient failure broke all subsequent tests until the process was restarted.

#### Error handling improvements:

* Added two private helpers in `TmuExtensions.cs`:
  * `ReleaseTmuResources` - releases the assigned TMU contexts by calling `ClearAssignedTMUContexts` with `doTMUReleaseCheck: false`. It is best-effort: it issues no driver calls and suppresses any failure so the original exception is never masked.
  * `DoWithTmuReleaseOnFailure` (an `Action` and a `Func<TResult>` overload) - runs a TMU operation, and on failure calls `ReleaseTmuResources` before rethrowing the original exception.
* Routed every public TMU extension method through the wrapper: the lifecycle methods (`TMUInitiate`, `EnableTMU`, `TMUAbort`, `FetchAveragedTMUMeasurement`), all measurement configuration methods, all granular `ConfigureTMU*` setters, and all `GetTMU*` getters that use an assigned TMU.
* Moved all validation calls inside the wrapper. This is deliberate -  assignment happens up front, so an invalid argument passed to a later call must still release the TMUs assigned earlier.
* Gave `DisableTMU`, `AssignTMUResources` and `ClearTMUAssignment` a direct try/catch that calls `ReleaseTmuResources` and rethrows, rather than routing them through the wrapper, since the release is part of their own cleanup path.
* Left `GetTMUCount` unwrapped - it reads a session-level capability and works with no TMU assigned.

The TMU is intentionally not disabled during cleanup. Only the assignment is released; the driver handles unreserving a TMU that is still enabled at the hardware level.

### Testing
Validated the same in hardware with multiple consecutive runs and the TMU is being cleared after each of the run and all the tmu are available for `TmuContextManager` to aqquire.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -42,15 +42,18 @@ public static class TmuExtensions
         /// </exception>
         public static void TMUInitiate(this DigitalSessionsBundle sessionsBundle, string[] pinNames = null)
         {
-            ValidatePinsOfTMU(sessionsBundle.Pins, pinNames);
-            sessionsBundle.Do((sessionInfo, sitePinInfo) =>
+            sessionsBundle.DoWithTmuReleaseOnFailure(pinNames, () =>
```
NikhilSivadas-NIB, line 45: Same comment as in the AzDO PR. Could we add a unit test covering a multithreaded scenario that could create a race condition?

NikhilSivadas-NIB (commented):
Partially reviewed but can you look at this comment?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: As this is difficult to reproduce form the ATS, I have added the steps in the existing manual test to run the TMU sequence example for 20 or more lots.
More info : https://dev.azure.com/ni/DevCentral/_git/srd-sebu-mixed-signal-apt/pullRequest/1383622?discussionId=10653479#1788763654

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -1734,6 +1861,71 @@ private static bool IsSafeToReleaseAllTMUs(NIDigital session, IEnumerable<SitePi
...
+        /// <param name="operation">The TMU operation to invoke.</param>
+        private static void DoWithTmuReleaseOnFailure(this DigitalSessionsBundle sessionsBundle, string[] pinNames, Action operation)
+        {
+            try
+            {
+                operation();
+            }
+            catch
+            {
+                ReleaseTmuResources(sessionsBundle, pinNames);
```
NikhilSivadas-NIB, line 1934: Right now on Enable TMU we set  tmu.Enabled = true, its at driver level property right ?  And following that as per new implementation any operation that throw excerption , the exception handler is invoking ReleaseTmuResources with internally calling ClearAssignedTMUContexts which removes data from TMUContextManager, but it does not make the tmu.Enabled= false  right? 
Do we need to make tmu.Enabled = false as well as part of exception handling? 

- Otherwise next time AssignTMUContexts method getting the disabled tmu from driver right? will it list the available TMUs that are not marked tmu.Enabled= false  ?
`List<string> availableTMUContexts = GetDigitalTmus(digitalSessionInformation.Session).GetDisabledTmuContexts();`


Are you facing any such issue?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -1413,10 +1504,13 @@ public static PinSiteData<TmuSourceEvent> GetTMUStartSourceEvent(this DigitalSes
         /// <exception cref="NISemiconductorTestException">Thrown when a TMU resource has not been assigned to one or more pins. Call <see cref="AssignTMUResources(DigitalSessionsBundle, string[])"/> before invoking this method.</exception>
         public static PinSiteData<TmuSourceEvent> GetTMUStopSourceEvent(this DigitalSessionsBundle sessionsBundle)
         {
-            return sessionsBundle.DoAndReturnPerSitePerPinResults((sessionInfo, sitePinInfo) =>
+            return sessionsBundle.DoWithTmuReleaseOnFailure(null, () =>
```
NikhilSivadas-NIB, line 1507: just to confirm, in case of one thread per site in TestStand, if there is a exception during execution of any GetXX methods within DoWithTmuReleaseOnFailure, then ReleaseTmuResources will only ClearAssignedTMUContexts for that particular site right? So that other threads for other sites could still continue execution

NikhilSivadas-NIB (commented):
can you check the comment and reply?

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Updated the code to disable the tums. @NikhilSivadas-NIB Please have a look.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
Ashutosh-NI: Yes, that's correct. Each thread maintains its own bundle, and in `DoWithTmuReleaseOnFailure` we only clear the bundle associated with that specific thread. This ensures that cleanup is isolated to the failing thread and does not impact bundles managed by other threads.

NikhilSivadas-NIB (approved):
Approved

PR #605, opened 2026-09-02
ForceV(or I)Sequence to always wait for source completion

nitamil:

Updates DCPower sequence-forcing APIs so they wait for sequence completion by default.

- Changes `ForceVoltageSequence` to always wait for source completion.
- Changes `ForceCurrentSequence` so `waitForSequenceCompletion` defaults to `true`.
- Removes the completion-wait parameter from the core sequence implementation and always initiates channels with completion waiting enabled.
- Adds new `ForceVoltageSequence` overloads without the completion-wait parameter.
- Marks the existing overloads that accept `waitForSequenceCompletion` as obsolete to guide callers toward the new API.

Callers should not continue execution until a forced voltage or current sequence has completed. Making this behavior the default prevents race conditions caused by code proceeding before the instrument finishes applying the requested sequence.

The obsolete overloads preserve source compatibility while providing a migration path to the simplified APIs.

Existing Auto tests are passing after the removal of the parameter.


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -342,14 +342,50 @@ public static void ForceVoltage(this DCPowerSessionsBundle sessionsBundle, PinSi
...
         /// <param name="sequenceTimeoutInSeconds">Maximum time to wait for completion when <paramref name="waitForSequenceCompletion"/> is <see langword="true"/>.</param>
+        [Obsolete("This method has been deprecated. Use the ForceVoltageSequence() overload without the waitForSequenceCompletion parameter instead.")]
+        public static void ForceVoltageSequence(
+            this DCPowerSessionsBundle sessionsBundle,
+            double[] voltageSequence,
+            double? currentLimit = null,
+            double? voltageLevelRange = null,
+            double? currentLimitRange = null,
+            int sequenceLoopCount = 1,
+            bool waitForSequenceCompletion,
```
Ashutosh-NI: `waitForSequenceCompletion` Should be have default false value. I would suggest to have original parameter list and only the body should be changed to redirect it to new function.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -446,6 +482,56 @@ public static void ForceVoltageSequence(
             sessionsBundle.ReleaseAdvancedSequenceResources(advancedSequenceName);
         }
```
Ashutosh-NI, line 485: Please using the newly created function and redirect this to use that instead of having same logic in both of the places.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -461,6 +547,7 @@ public static void ForceVoltageSequence(
         /// <param name="sequenceLoopCount"/>
         /// <param name="waitForSequenceCompletion"/>
         /// <param name="sequenceTimeoutInSeconds"/>
+        [Obsolete("This method has been deprecated. Use the ForceVoltageSequence() overload without the waitForSequenceCompletion parameter instead.")]
```
Ashutosh-NI, line 551: Please using the newly created function and redirect this to use that instead of having same logic in both of the places.

SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/DCPower/Source.cs
```diff
@@ -3829,7 +3966,7 @@ private static void ForceSequenceCore(
                 outputFunction: (DCPowerSourceOutputFunction)settings.OutputFunction,
                 setAsActiveSequence: setAsActiveSequence);
 
-            channelOutput.InitiateChannels(waitForSequenceCompletion, sequenceTimeoutInSeconds);
+            channelOutput.InitiateChannels(true, sequenceTimeoutInSeconds);
```
Ashutosh-NI: I think we might need to do the similar (i.e. making the old function obsolete and adding the new function instead without `waitForSequenceCompletion` parameter), please make the required changes.

Ashutosh-NI (changes requested):
Can we add the auto test for the new function or update the existing to test methods as well?

PR #609, opened 2026-09-09
Examples | Digital | SPI Protocol

Shalini-Subramanian:
This pull request introduces a protocol abstraction layer for digital register IO operations, laying the groundwork for supporting multiple digital communication protocols (such as SPI ) in the DUT control code. The main changes include the definition of a protocol-agnostic interface, an abstract base class for protocol implementations, a new bit order enumeration, and an extensible protocol selector.

This PR contains the implementation for the SPI example

**Key changes:**

### Digital Protocol Abstraction

* Added the `IDigitalProtocol` interface in `IDigitalProtocol.cs`, defining a protocol-agnostic contract for register read/write operations, waveform management, and configuration (including bit order, pin names, and default widths). This interface enables consistent interaction with different protocol implementations.
* Introduced the `BitOrder` enum to specify MSB-first or LSB-first sample ordering, supporting flexible waveform packing/unpacking for various protocols.
* Added the `CommunicationProtocol` enum for identifying and selecting the protocol implementation at runtime (currently only SPI is listed).

### Base Implementation and Utilities

* Added an abstract `DigitalProtocol` base class in `DigitalProtocol.cs` that implements most of the logic for register access (read/write, single and multi-register), waveform creation, and sample conversion. This class centralizes common protocol logic, making it easier to implement specific protocols like SPI or I2C.
* Implemented robust sample conversion helpers and validation logic in the base class, ensuring reliable packing/unpacking of register values and validating pin configurations against the instrument bundle.
TODO: Justify why this contribution should be part of the project.

- To be added


Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/DutControl/DigitalProtocol.cs
```diff
@@ -0,0 +1,377 @@
...
+        public string ReadPatternName { get; set; }
+
+        /// <inheritdoc/>
+        public string SourceWaveformName { get; set; } = "source_buffer";
+
+        /// <inheritdoc/>
+        public string CaptureWaveformName { get; set; } = "capture_buffer";
+
+        /// <summary>The pin that drives outgoing (source) data. For SPI this is SDI (MOSI).</summary>
+        public string SourcePinName { get; set; } = "SDI";
```
ni-ramya: `ConfigureDigitalProtocol` allows callers to override pinNames, but DigitalProtocol.SetBundle still creates waveforms using the hard-coded `SourcePinName = "SDI"`  and `CapturePinName = "SDO"`.

Either expose/configure `SourcePinName` and `CapturePinName`, or remove the independent `PinNames` option and derive validation/waveform pins consistently from the configured protocol pins.

Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/DutControl/DigitalProtocol.cs
```diff
@@ -0,0 +1,377 @@
...
+            {
+                // Create the source waveform as SiteUnique so both the broadcast-style writes (same data to
+                // every site) and the site-unique writes share one waveform name. A Broadcast waveform cannot
+                // be written with WriteSourceWaveformSiteUnique, whereas a SiteUnique waveform supports both.
+                _digitalSessionsBundle.CreateSerialSourceWaveform(
+                    SourcePinName, SourceWaveformName, NationalInstruments.ModularInstruments.NIDigital.SourceDataMapping.SiteUnique, SampleWidth, bitOrder);
+                _digitalSessionsBundle.CreateSerialCaptureWaveform(
+                    CapturePinName, CaptureWaveformName, SampleWidth, bitOrder);
+            }
+            catch (Exception)
```
ni-ramya: The catch block treats every exception as “waveforms already exist.” That also hides invalid pin names, invalid sample widths, unsupported bit orders, instrument/session failures, and other real initialization errors. The subsequent burst then fails at a less useful location—or can operate against stale waveforms.

If the driver exposes a specific exception/status for an existing waveform, catch only that condition. Otherwise, explicitly query for existence before creating the waveform and allow all other failures to propagate.

Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/TestSteps/ConfigureDigitalProtocol.cs
```diff
@@ -0,0 +1,63 @@
...
+            string readPatternName = "SPI_read_template",
+            string sourceWaveformName = "source_buffer",
+            string captureWaveformName = "capture_buffer",
+            int sampleWidth = 8,
+            string readWriteCountSequenceRegister = "reg0",
+            string addressBitWidthSequenceRegister = "reg1",
+            string valueBitWidthSequenceRegister = "reg2",
+            string[] pinNames = null)
+        {
+            IDigitalProtocol protocol = SPI.Instance;
```
ni-ramya: There us a protocol selector in the constructor but here we always create a SPI instance.

Examples/tests/Unit/Sequence/RegisterIO.SPI-I2C/TestSteps/ConfigureDigitalProtocolTests.cs
```diff
@@ -0,0 +1,93 @@
...
+using Xunit;
+
+namespace NationalInstruments.Examples.SemiconductorTestLibrary.RegisterIO.SPIAndI2C.Test.TestSteps
+{
+    public class ConfigureDigitalProtocolSPITests
+    {
+        [Fact]
+        public void ConfigureDigitalProtocol_AddressBitWidth_UpdatesSpiInstance()
+        {
+            TestStep.ConfigureDigitalProtocol(CommunicationProtocol.SPI, addressBitWidth: 8);
```
ni-ramya: **[Medium] ConfigureDigitalProtocolTests.cs — Tests mutate shared singleton state**
These tests modify SPI.Instance but do not restore its defaults. Other tests assert the default values, so results depend on test order and parallel execution. Use a fresh SPI instance per test, reset the singleton in teardown, or place the tests in a non-parallel xUnit collection.

It is primarily an issue in the unit tests, specifically the interaction between:

1. SPI.Instance in SPI.cs, which exposes mutable global state.
2. ConfigureDigitalProtocolTests, which changes that state.
3. SpiDefaultParametersTests, which expects the original defaults.

Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/DutControl/DigitalProtocol.cs
```diff
@@ -0,0 +1,377 @@
...
+        /// <inheritdoc/>
+        public SiteData<long> ReadRegister(uint address)
+        {
+            return ReadRegister(address, DefaultAddressBitWidth, DefaultValueBitWidth);
+        }
+
+        /// <inheritdoc cref="ReadRegister(uint)"/>
+        /// <param name="addressBitWidth">The bit width for the register address.</param>
+        /// <param name="valueBitWidth">The bit width for the register value.</param>
+        public SiteData<long> ReadRegister(uint address, uint addressBitWidth, uint valueBitWidth)
```
ni-ramya: Several public methods assume valid, matching inputs:

1. WriteRegisters(uint[] addresses, long[] values) assumes equal lengths.
2. WriteRegisters(uint[] addresses, SiteData<long>[] values) assumes values is non-empty and all SiteData instances contain identical site sets.
3. ReadRegisters and U32SamplesToLongArray do not handle empty address lists.
4. GetSampleCount divides by sampleBitWidth, so sampleWidth == 0 causes a divide-by-zero exception.
5. Values larger than the selected width are silently truncated.
6. Widths greater than 64 bits can result in invalid shifts against long.

At minimum, add argument validation
```
+ if (sampleWidth == 0 || sampleWidth > 32)
+     throw new ArgumentOutOfRangeException(nameof(sampleWidth));
+
+ if (addresses == null || values == null)
+     throw new ArgumentNullException(...);
+
+ if (addresses.Length != values.Length)
+     throw new ArgumentException("Addresses and values must have equal lengths.");
```

ni-ramya: Also add tests for invalid widths, empty arrays, mismatched lengths, and values that exceed the configured width.

Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/DutControl/SPI.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// SiteData&lt;long&gt; readBack = protocol.ReadRegister(address: 0x48);
+    /// </code>
+    /// </para>
+    /// </remarks>
+    public sealed class SPI : DigitalProtocol
+    {
+        private static readonly Lazy<SPI> _instance = new Lazy<SPI>(() => new SPI());
+
+        /// <summary>Gets the singleton SPI protocol instance.</summary>
+        public static SPI Instance => _instance.Value;
```
ni-ramya: SPI.Instance is a process-wide mutable singleton, while SetBundle stores the active DigitalSessionsBundle in an instance field:

Every call to DutControl rebinds the singleton to a new bundle and shares all configuration globally. This is unsafe if:

* multiple test programs run in the same process,
* sites or contexts are used concurrently,
* tests execute in parallel,
* one test changes configuration after another has configured the singleton.

Prefer constructing a protocol instance per module/context, or separate immutable protocol configuration from the per-session operation object. If the singleton is retained for example simplicity, the documentation should explicitly state that it is not thread-safe and the tests should cover repeated context initialization.

Examples/tests/Integration/Sequence/RegisterIO.SPI-I2C/WriteSiteUniqueValueToRegisterAndCompareReadbackValuesTests.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace NationalInstruments.Tests.SemiconductorTestLibrary.Integration
+{
+    [Collection("NonParallelizable")]
+    public class WriteSiteUniqueValueToRegisterAndCompareReadbackValuesTests
+    {
+        private const string PinMapFileName = "STLExample.RegisterIO.SPIAndI2C.pinmap";
+        private const string SupportingMaterialsFolderPath = @"\Examples\source\Sequence\RegisterIO.SPI-I2C\Supporting Materials\DigitalProject";
+        private const string DigitalPatternProjectName = "STLExample.RegisterIO.SPIAndI2C.digiproj";
+
+        [Theory(Skip = "Requires the shared digital project and a digital pattern instrument.")]
```
ni-ramya: Is it possible to have this executing with a test mock?

Examples/tests/Tests.SemiconductorTestLibrary.Examples.csproj
ni-ramya: Will there be a readme file like other examples? Are we palnning to taek it up in thir PR or sparate? Make sure to include

1. Required STS/software versions
2. Required digital pattern instrument and DUT wiring
3. Pin names and their SPI roles
4. How to open and run the sequence
5. How to build the code module
6. Required pin map, digital project, timing, levels, and specs files
7. How to change address/value widths and waveform names
8. Expected readback behavior
9. Offline-mode limitation

Examples/source/Sequence/RegisterIO.SPI-I2C/Code Modules/DutControl/CommunicationProtocol.cs
```diff
@@ -0,0 +1,15 @@
+namespace NationalInstruments.Examples.SemiconductorTestLibrary.RegisterIO.SPIAndI2C.DutControl
+{
+    /// <summary>
+    /// Identifies the digital communication protocol used to read and write DUT registers.
+    /// The <c>TestStep</c> methods use this value to select which <see cref="IDigitalProtocol"/>
+    /// implementation to instantiate at runtime.
+    /// </summary>
+    public enum CommunicationProtocol
```
ni-ramya: I would suggest naming this as `CommmunicationProtocolType` or `DigitalProtocolType` to be more explicit.

Examples/tests/Integration/Sequence/RegisterIO.SPI-I2C/WriteSiteUniqueValueToRegisterAndCompareReadbackValuesTests.cs
ni-ramya: If it is going to stay offline, add a section in the README explaining how to get it running

ni-ramya (changes requested):

Examples/tests/Integration/Sequence/RegisterIO.SPI-I2C/WriteSiteUniqueValueToRegisterAndCompareReadbackValuesTests.cs
Shalini-Subramanian: I will get back

Shalini-Subramanian:
@Mattjet27 ,could you please share your thoughts on the high-level structure and hierarchy of the SPI protocol example? This will help me have it ready by the time you’re back.

PR #612, opened 2026-09-11
Add HintPath validation for Example projects

manshita07:

The purpose is to catch broken project references early and produce actionable failures when an Example project contains an invalid dependency path. The validation is intentionally narrow and only checks the actual path resolution problem.

This change helps prevent Example projects from silently breaking due to stale or incorrect HintPath values. Broken HintPaths can cause confusing runtime or build failures that are difficult to diagnose, especially in shared Example content. By validating the resolved path at test time, we catch these issues earlier and make the failing dependency easier to identify.

This validation was exercised through the Example project test suite in ATS. The test was run against the Example projects and the result was reviewed from the ATS execution output to confirm pass/fail behavior for actual HintPath resolution issues.

image


Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
+﻿﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Xunit;
+namespace NationalInstruments.Tests.SemiconductorTestLibrary.Functionality.HintPathValidation
```
hcnagaraj-e: Change the namespace `NationalInstruments.Tests.SemiconductorTestLibrary.Functionality.Example`

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    public class ValidateExamplesHintPath
+    {
+        public static IEnumerable<object[]> GetExampleProjectPaths()
+        {
+            string FindExamplesRoot()
```
hcnagaraj-e: You do not need all these things; example directory will be guaranteed in that place.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+                    }
+
+                    current = current.Parent;
+                }
+
+                throw new DirectoryNotFoundException("Could not locate Examples root from current working directory.");
+            }
+
+            var examplesRoot = FindExamplesRoot();
+            var projectsRoot = Path.Combine(examplesRoot, "source");
```
hcnagaraj-e: Update this step to get source directly by hardcoding the relative pathd

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+                throw new DirectoryNotFoundException(string.Format("Examples source root not found: {0}", projectsRoot));
+            }
+
+            foreach (var projectPath in Directory.GetFiles(projectsRoot, "*.csproj", SearchOption.AllDirectories))
+            {
+                yield return new object[] { projectPath };
+            }
+        }
+
+        [Theory]
```
hcnagaraj-e: Check if you can use theory data instead. @Ashutosh-NI What would be the best option to repeat the test for each project?

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+        [MemberData(nameof(GetExampleProjectPaths))]
+        public void ValidateExamplesHintPaths_WhenPathsAreMissing_ShouldFailPerExample(string projectPath)
+        {
+            var issues = ValidateExampleHintPaths(projectPath);
```
hcnagaraj-e: Break this big function in to different stages, you can use helper methods to simplifies each stage.

Ex:
Stage1. Get all hint paths as List
Stage2. Validate `HintPaths` in a loop and create another List of `BadHintPaths`
Stage3. Assert bad hint path array/list. (No need to do extra formatting, let's see if we need to improve the reporting)

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+            Assert.True(issues.Count == 0, BuildIssueReport(projectPath, issues));
+        }
+
+        private static List<HintPathIssue> ValidateExampleHintPaths(string projectPath)
```
hcnagaraj-e: As mentioned in previous comment, break this and move it to the actual test method.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+
+                foreach (var hintNode in hintNodes)
+                {
+                    var rawHintPath = hintNode.Value.Trim();
+                    if (string.IsNullOrEmpty(rawHintPath))
+                    {
+                        continue;
+                    }
+
+                    var expandedHintPath = Environment.ExpandEnvironmentVariables(rawHintPath);
```
hcnagaraj-e: This should be moved to main test.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,144 @@
...
+                    if (!File.Exists(Path.GetFullPath(resolvedPath)))
+                    {
+                        AddIssue(referenceName, rawHintPath, "HintPath target file does not exist.");
+                    }
+                }
+            }
+
+            return issues;
+        }
+        private static string BuildIssueReport(string projectPath, IReadOnlyCollection<HintPathIssue> issues)
```
hcnagaraj-e: All these things not required.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Xml.Linq;
+using Xunit;
+
+namespace NationalInstruments.Tests.SemiconductorTestLibrary.Functionality.HintPathValidation
+{
+    public class ValidateExamplesHintPath
+    {
+        public static IEnumerable<object[]> GetExampleProjectPaths()
+        {
+            var current = new DirectoryInfo(Directory.GetCurrentDirectory());
```
hcnagaraj-e: Simplify this logic to directly get example source path.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                yield return new object[] { projectPath };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(GetExampleProjectPaths))]
+        public void ValidateExamplesHintPaths_WhenPathsAreMissing_ShouldReportInvalidHintPaths(string projectPath)
+        {
+            var invalidHintPaths = ValidateExampleHintPaths(projectPath);
```
hcnagaraj-e: Can you update the main tests as discussed in desk review.

I am summarizing it again here:
- Keep all the main steps in Test block itself, you can still have helper method, but it should not be deeper. The logic or the main steps should be visible in test itself, it should not be totally hidden behind the helper logic.
- Keep the logic simple to get example root directory simple. 
- Get all the project paths and ensure test is repeated for each project.
- In test logic, get the references and hint paths of the project under test
- Validate each hint path, report all invalid hint path
- Assert all invalid hint path with simple assert, no need to format into specific format.

hcnagaraj-e (changes requested):
Please fix all the comments or update your response before reset.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: _BuildIssueReport()_ has been removed and now simple List of invalid hintpaths is returned.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Changed the namespace namespace NationalInstruments.Tests.SemiconductorTestLibrary.Functionality.HintPathValidation to NationalInstruments.Tests.SemiconductorTestLibrary.Functionality.Example

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Replaced  FindExamplesRoot() with hardcoded relative path to source folder

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Added harcoded relative path to get the source directly.
 var sourceFolderPath = Path.GetFullPath(Path.Combine(
                Directory.GetCurrentDirectory(),
                "..",
                "..",
                "..",
                "..",
                "source"));

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Created helper methods GetHintPaths() and GetInvalidHintPaths() to simplify the function into stages.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Added to the main test after the hintpaths are validated.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: - [TheoryData<string] is used in [ValidateExamplesHintPath.cs] to hold all project paths as strongly typed test data.
-The test is repeated for each project by using [[Theory]] with [[MemberData(nameof(GetExampleProjectPaths))]] in [ValidateExamplesHintPath.cs].
- The example root lookup is kept simple by building the source path from a direct relative path.
- The test block still shows the main flow:
     load [projectXml]→ get [hintPaths]→ validate invalid hint paths → assert.
- Invalid hint paths are collected into a list and then asserted together.
- The assert is simple and does not use any custom formatting structure like CSV.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: TheoryData is used to hold all the project paths and each project path is fetched by MemberData to repeat the test.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: The actual test now calls GetInvalidHintPaths() method which simplifies and increases code readability.

Examples/tests/ValidationTests/ValidateExamplesHintPath.cs
manshita07: Hardcoded the relative source path to directly get it.

PR #613, opened 2026-09-11
Add Examples to Release Asset 26.0.0+

manshita07:

This PR updates the release draft workflow to include shipped Examples as a release asset starting from version 26.0.0, while preserving behavior for earlier versions.

This change implements the release improvement requirement to publish shipped Examples for 26.0.0+ releases and keeps release automation backward-compatible.

Why this matters:

Meets the new release expectation: examples are part of customer-facing release assets from 26.0.0 onward.
Protects older release paths: pre-26 versions continue publishing valid assets (.nupkg and Assemblies.zip) without breaking on missing Examples.zip.
Improves workflow clarity and maintainability with explicit conditional release paths and user-facing skip messaging.

End-to-end artifact generation check for 26.0.0 via the connected AzDO artifact flow:
   - Examples export is resolved and downloaded.
   - SemiconductorTestLibrary.26.0.0.Examples.zip is created successfully.
   - Artifacts are staged successfully for GitHub upload.
   
Negative-path behavior validated by design:
   - For versions below 26.0.0, workflow proceeds without expecting Examples.zip and logs an explicit notice.


.github/workflows/draft-release-notes.yml
```diff
@@ -189,9 +189,32 @@ jobs:
...
+          if [ "$(printf '%s\n' "$VERSION" "26.0.0" | sort -V | head -n1)" = "26.0.0" ]; then
+            mv jfrog-artifacts/SemiconductorTestLibrary.$VERSION.Examples.zip .
+            echo "INCLUDE_EXAMPLES=true" >> $GITHUB_ENV
+          else
+            echo "::notice::Version $VERSION is below 26.0.0; Examples.zip is not expected. Proceeding with .nupkg and Assemblies.zip only."
+            echo "INCLUDE_EXAMPLES=false" >> $GITHUB_ENV
+          fi
       # Create Release Notes Draft
-      - name: Create Release Notes Draft
+      - name: Create Release Notes Draft (With Examples)
```
Utkarsh-NI-Emerson, line 200: Why are we adding another step for creating Release Notes draft "with examples", it should be directly included in the earlier step that is now changed to "without examples".
I don't think there will be a scenario where we do not want to include examples.

.github/workflows/draft-release-notes.yml
```diff
@@ -189,9 +189,32 @@ jobs:
           # Move the expected files to the current directory (they are in flat layout already)
           mv jfrog-artifacts/NationalInstruments.SemiconductorTestLibrary.$VERSION.nupkg .
           mv jfrog-artifacts/SemiconductorTestLibrary.$VERSION.Assemblies.zip .
-
+          if [ "$(printf '%s\n' "$VERSION" "26.0.0" | sort -V | head -n1)" = "26.0.0" ]; then
```
Utkarsh-NI-Emerson, line 192: Why are we focusing on the release version 26.0.0?
This goes same for the AzDO PR.

Utkarsh-NI-Emerson:
Please attach validation links and screenshots.

Utkarsh-NI-Emerson:
Please attach the corresponding AzDO PR to the description as this change won't make any sense without a context of the other.

Utkarsh-NI-Emerson (changes requested):

.github/workflows/draft-release-notes.yml
manshita07: Earlier I divided create Release Notes draft into with/without examples taking into consideration that examples are to be included only for version >= 26.0.0 
Removed the step for creating Release Notes draft with examples and included it directly under the earlier step.

.github/workflows/draft-release-notes.yml
manshita07: I interpreted that examples zip is to be added for versions 26.0.0 and onwards and hence created conditions following that in both Github and AzDO PR.

PR #615, opened 2026-09-16
Fix TMU cleanup

Ashutosh-NI:
This pull request improves TMU extension methods in the `TmuExtensions.cs` file. The main changes ensure that TMU resources are always released and properly cleaned up if an exception occurs, regardless of which pins were targeted by the operation. Additionally, the XML documentation for each public method is updated to clearly describe this behavior, making the API more predictable and safer to use.

The most important changes are:

**Exception Safety and Resource Management:**
- All TMU-related extension methods now guarantee that if an exception is thrown, TMU resources assigned to every pin in the `DigitalSessionsBundle` are disabled and released back to the resource pool, not just those targeted by the method. This is achieved by removing the `pinNames` argument from calls to `DoWithTmuReleaseOnFailure` and `ReleaseTmuResources`, ensuring bundle-wide cleanup. 

**Documentation Improvements:**
- Updated the XML documentation for all relevant methods to explicitly state that if an exception occurs, all TMU resources in the bundle are released and must be reassigned before retrying. This clarifies the contract for users of the API. 

## Testing

Ran the example with updated code and validated its working in hardware.
image

ATS Run - [Offline ATS](https://dev.azure.com/ni/DevCentral/_build/results?buildId=21848484&view=results)


SemiconductorTestLibrary.Extensions/source/InstrumentAbstraction/Digital/TMU/TmuExtensions.cs
```diff
@@ -1865,20 +1959,23 @@ private static bool IsSafeToReleaseAllTMUs(NIDigital session, IEnumerable<SitePi
...
         {
             try
             {
                 sessionsBundle.Do((sessionInfo, sitePinInfo) =>
                 {
                     // A pin may have no assigned TMU context, since the failure being cleaned up
                     // after may have occurred part way through assigning TMU resources.
-                    if (!DoForThisPin(pinNames, sitePinInfo.PinName)
-                        || string.IsNullOrEmpty((sitePinInfo as DigitalSitePinInfo)?.AssignedTmuContext))
+                    if (string.IsNullOrEmpty((sitePinInfo as DigitalSitePinInfo)?.AssignedTmuContext))
```
Ashutosh-NI, line 1975: @NikhilSivadas-NIB 
**`ReleaseTmuResources` no longer filters by `pinNames` (removed `DoForThisPin`)**

TMU assignment is tracked **per pin, for the lifetime of the process**. Cleanup-after-failure is *bundle* cleanup, not *pin* cleanup - once a step throws, the caller abandons the whole bundle (means the reference will never be used again), so any pin left assigned is leaked forever.

Filtering the release by `pinNames` only released the pins the **failed operation** targeted, which strands every other pin in the bundle.

---

###  failure after pins are already enabled

```csharp
bundle.AssignTMUResources();                                       // C0, C1 assigned
bundle.ConfigurePeriodMeasurement(...);                            // =C0, C1 now Enabled = true
bundle.ConfigureTMUSkewMeasurement(new[]{ "NonExistentPin" }, ...); //  validation throws
```

- **Old:** the release is scoped to `{ "NonExistentPin" }`, which matches **no pin in the bundle**, so *nothing* is released. `C0` and `C1` stay `Enabled = true` and stay assigned — the TMUs remain reserved in hardware and unavailable to every later assignment.
- **New:** `C0` and `C1` are disabled and released, so the process returns to a clean state.

---

### Impact

Each stranded pin permanently shrinks the TMU pool. After enough failures, unrelated tests start failing with "no TMU available" - which is exactly the cascading failure pattern we saw in the test run.

---

### Regression coverage

`ConfigureFailsWithPinNotInBundle_TMUAssignmentIsReleasedForEveryPinInBundle` asserts `AssignedTmuContext` is empty for **every** pin in `AggregateSitePinList` after the failure. It fails on the old behavior regardless of how large the TMU pool is.

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
```diff
@@ -296,8 +294,22 @@ void ConfigurePeriodMeasurementWithEitherEdge()
 
             Assert.Throws<NISemiconductorTestException>(() =>
                 ConfigurePeriodMeasurementWithEitherEdge());
-            sessionsBundle.DisableTMU();
-            sessionsBundle.ClearTMUAssignment();
+        }
+
+        [Fact]
+        public void ConfigureFailsWithPinNotInBundle_TMUAssignmentIsReleasedForEveryPinInBundle()
```
NikhilSivadas-NIB: Can we rename the test with AAA for better readability? This covers a particular behavior of ConfigurePeriodMeasurement right by passing invalid pin? And we must Assert.Throws Exception and also clear TMU,

NikhilSivadas-NIB: And also, we are changing tmu.Enabled to false along with ReleasingTMU from Context manager, is it possible to include that as well as part of Assert ?

NikhilSivadas-NIB: It's a documented behavior now, so shall we add the test case for these behavior for all the applicable method which stick to this behavior?

NikhilSivadas-NIB (commented):

SemiconductorTestLibrary.Extensions/tests/Unit/InstrumentAbstraction/Digital/TMU/TmuExtensionsTests.cs
Ashutosh-NI: Updated the name to have AAA format.

Ashutosh-NI: Added a helper function which will assert both disable of tmu and cleared the assigned tmus.

Ashutosh-NI: Added a helper function which will assert both disable of tmu and cleared the assigned tmus.

NikhilSivadas-NIB (approved):
Approved

